PR #1, opened 2020-01-10
Adding design proposal for generic miniRA interface.

cfallin:
This is a specific implementation proposal for now, but could morph into
a more complete piece of API documentation if/when we expose this as a
crate for anyone to use.

### DIFF
--- a/docs/interface-proposal.md
+++ b/docs/interface-proposal.md
@@ -1,0 +1,488 @@
+Register allocator: generic interface
+=====================================
+
+This document describes the plan to build a generic interface to the
+new register allocator in the `minira` repository, to be used by the
+new instruction selection backend in Cranelift. The end goal is that
+the register allocators live in their own crate with a well-defined,
+generic interface, and Cranelift can import and use the crate by
+providing the appropriate data and implementing the appropriate traits
+on its machine-instruction data structures.
+
+This design originated with, and is still largely due to, the
+ideas/design from Julian Seward (email on 2020-01-08), and has since
+been updated to add more detail, describe concrete Rust definitions,
+etc.
+
+A few design decisions should be highlighted. First, the input to the
+register allocator is a list of machine instructions with *virtual
+registers*. Importantly, this is *not SSA*. In other words, a
+particular virtual register may be defined in several places. It is
+expected that the client (rest of compiler backend) will lower any
+preceding SSA IR's phi-nodes into explicit moves beforehand. (The
+register allocator will often be able to eliminate these moves, so it
+is alright to initially use a lowering strategy that
+naively/unconditionally inserts many moves.)
+
+Second, the register allocator does not track or manage stack layout,
+so its spill slots are abstractions.
+
+Finally, the register allocator is designed as a *pure computation*:
+the input function is immutable, and the output is a new sequence of
+instructions along with some metadata (like which registers were
+actually used). This is likely to be nearly as efficient as a
+mutate-in-place approach, if not more so, because the allocator will
+need to splice many instructions into the function body in any case;
+if it performs this rewrite in one final pass, writing a new sequence
+of instructions as it goes, it might as well return the new sequence
+separately. Incidentally, taking a completely immutable &Function
+(rather than &mut Function) is nice in the future if we want to do
+things like parallel codegen -- slightly lower friction in idiomatic
+Rust.
+
+Register file descriptions
+--------------------------
+
+The client of the register allocator provides information about the
+register file(s). We will first describe the logical view of the world
+as the regalloc sees it, and then how the resulting concepts are
+encoded compactly.
+
+The register allocator allocates registers from one or more *register
+classes*. There are expected to be only a small number of classes on a
+given machine: e.g., all integer registers and all floating-point
+registers. Unlike some other compiler designs, we do not expect to
+define special register classes for, e.g., overlapping narrow subsets
+of registers (low half or low byte registers), or subsets of the
+register file usable by certain instructions; overlapping classes
+complicate the allocator's job. Instead, we have a few major, disjoint
+classes, and deal with instruction-specific constraints by simply
+specifying a specific real register at instruction selection time.
+
+In our proposed encoding, up to 8 classes (chosen by 3 bits) can be
+supported. As an example, on ARM64, there will initially be a GPR
+class (I64, x0..x30 minus special registers) and an FP/vector class
+(F64 / V128, v0..v31). x86-64 is similar: the GPR class is
+rax/.../r8..r15 and the FP/vector class is xmm0..xmm15.
+
+All registers, in all register classes, live in one contiguous index
+space. Each class corresponds to a range within this index space.
+
+Thus, the machine backend provides to the regalloc the following
+information (the name `RRegUniverse` is borrowed from Valgrind, which
+uses the same data structure):
+
+    struct RRegUniverse {
+        num_regs: RReg,
+
+        /// `ranges` is indexed by register class and provides the
+        /// (start, end) ranges in the contiguous index space for each range.
+        /// The registers in that class are numbered `start` to `end-1`,
+        /// inclusive.
+        ranges: Vec<(RReg, RReg)>,  // indexed by RegClass.
+    }
+        
+We can refer to a physical (real) register using the `RReg` type,
+which is just a newtype wrapper around the integer index:
+
+    struct RReg(u32);
+    
+    struct RegClass(u8);  // Actually more constrained (3 bits).
+    
+Identifying Real and Virtual Registers
+--------------------------------------
+
+First, the input to the register allocator may refer to virtual
+registers. There are an effectively unlimited number of virtual
+registers (limited only by the number of bits we provide for the index
+in our data structures), mapped to the much smaller number of real
+registers during allocation.
+
+However, while emitting instructions to provide to the regalloc, we
+often need to refer to not just virtual registers but also real
+registers: certain constraints (due to the ISA or the ABI) require
+certain values to be placed in particular registers, specified during
+instruction selection and respected by the regalloc.
+
+Thus, each instruction that refers to a register should use a union
+type that can name either a virtual register index or a real register
+index.
+
+In addition, every register reference indicates the associated
+register class, so that the allocator can easily know the associated
+constraints of a virtual register without referring to a separate
+constraint data structure.
+
+So we have, *conceptually*, the following data for every register slot
+in a machine instruction:
+
+    enum Reg {    // Not the real (memory-efficient) implementation.
+      RReg(RegClass, RReg),
+      VReg(RegClass, VReg),
+    }
+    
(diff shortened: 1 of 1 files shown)


bnjbvr:
Thanks! Taking as is, we can iterate and open issues for remaining discussions.

docs/interface-proposal.md
```diff
@@ -0,0 +1,482 @@
...
+is provided with two functions in the `Function` trait.
+
+    struct InsnRegUses {
+        used: RegSet,      // registers that are read.
+        defined: RegSet,   // registers that are written.
+        modified: RegSet,  // registers that are both read and written.
+                           // Note that `modified` is distinct from
+                           // just `used`+`defined` because the vreg
+                           // must live in the same real reg both
+                           // before and after the instruction.
```
sunfishcode, line 220: If a register is in the "modified" set, is is also expected to be in the "used" and "defined" sets?

Also, "modified" doesn't connote being read. A common term for this construct is "tied" -- conceptually an input is tied to an output if they must be in the same register. Would that term work here?

docs/interface-proposal.md
```diff
@@ -0,0 +1,482 @@
...
+less indirection.
+
+Interface from RA to instructions: query registers
+--------------------------------------------------
+
+Next, the client must provide the ability for the regalloc to observe
+register slots in instructions and to mutate them. This functionality
+is provided with two functions in the `Function` trait.
+
+    struct InsnRegUses {
```
sunfishcode, line 213: Can you update the docs for this to reflect changes in the code?

docs/interface-proposal.md
```diff
@@ -0,0 +1,482 @@
...
+    // TODO: we may also want to consider a sparse data structure
+    // (HashMap<Reg, Reg> for example) if warranted by data.
+    type RegMap = Vec<Reg>;
+
+The information for each instruction is provided as a few register
+sets; these containers are unordered collections of either virtual or
+physical registers.
+
+A note on semantics: uses are considered to happen prior to the
+instruction's effect, and defs are considered to happen after it. A
```
sunfishcode, line 262: Does the design have any provision for what is sometimes callled "early clobber" constraints, where an output is written to before the inputs are read? These also mostly occur in sequences where multiple ISA instructions are grouped together as one logical instruction.

docs/interface-proposal.md
```diff
@@ -0,0 +1,482 @@
...
+
+Implementation plan
+-------------------
+
+Once we have agreed on a design, Chris plans to build this abstraction
+layer into the `miniRA` repository while maintaining its existing
+concrete IR implementation (and interpreter and testing
+infrastructure). This should eventually result in a nicely abstracted
+register-allocator library, which can then be imported into Cranelift
+and used by the new instruction selection backend.
```
sunfishcode, line 482: This section can be updated now!

docs/interface-proposal.md
```diff
@@ -0,0 +1,482 @@
+Register allocator: generic interface
+=====================================
+
+This document describes the plan to build a generic interface to the
+new register allocator in the `minira` repository, to be used by the
```
sunfishcode, line 5: It's the `regalloc` repository now.

(Actually, it's the `regalloc.rs` repository, but the `regalloc` crate; is the `.rs` needed in the repository name?)

bnjbvr: Re naming, I thought that calling the repository `regalloc.rs` indicated Rust in addition to just "register allocation", which was more informative. Some other crates seem to do this. Of course not a big deal, so we could remove the file extension there.

docs/interface-proposal.md
```diff
@@ -0,0 +1,482 @@
...
+e.g. the I64 register class for GPRs and V128 for
+floating-point/vector registers.)
+
+Unlike some other compiler designs, we do not expect to define special
+register classes for, e.g., overlapping narrow subsets of registers
+(low half or low byte registers), or subsets of the register file
+usable by certain instructions; overlapping classes complicate the
+allocator's job. Instead, we have a few major, disjoint classes, and
+deal with instruction-specific constraints by simply specifying a
+specific real register at instruction selection time.
```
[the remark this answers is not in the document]
julian-seward1: I'll update it to point out the tradeoff.

There's no intention to support regclass subsets either, at least currently.  Maybe the RealRegUniverse structure could be extended at some point to support that, if needed.

I did think about the x86-32 byte-store problem, since that seems the most pressing case
(and at least for x86-* and arm* the only one I am aware of).  AFAICS there are no subset constraints for x86-64 or arm64.  As a short term kludge on x86-32 we could simply pretend that byte stores can be done also from %esi and %edi, and in such cases have the emitter emit (eg) `movb %edi, 1234(%eax, %edx, 4)` as `xchg %edi, %ecx; movb %ecx, 1234(%eax, %edx, 4) ; xchg %edi, %ecx`.  There will always be at least one low-half register available to swap with.  It's not ideal from a perf standpoint, but overall I'd be surprised if it really makes much difference overall.

docs/interface-proposal.md
```diff
@@ -0,0 +1,482 @@
...
+(v0..v31). x86-64 is similar: the I64 class is rax/.../r8..r15 and the
+V128 class is xmm0..xmm15.
+
+All real registers, in all register classes, live in one contiguous
+index space. Each class corresponds to a range within this index
+space.
+
+Thus, the machine backend provides to the regalloc the following
+information (the name `RRegUniverse` is borrowed from Valgrind, which
+uses the same data structure):
```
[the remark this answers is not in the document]
julian-seward1: We should remove all references to Valgrind; it's not necessary as a source of examples or motivation.  And use `RealReg` or "real register" instead of these short names.

docs/interface-proposal.md
```diff
@@ -0,0 +1,482 @@
...
+
+    struct RRegUniverse {
+        num_regs: RReg,
```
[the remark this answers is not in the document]
julian-seward1: It might have been at one time, but that field got removed in the implementation.  Now it just carries a `Vec` of regs and so carrying the number is obviously pointless.  I'll update the doc.

docs/interface-proposal.md
```diff
@@ -0,0 +1,482 @@
...
+
+        /// `ranges` is indexed by register class and provides the
+        /// (start, end) ranges in the contiguous index space for each range.
+        /// The registers in that class are numbered `start` to `end-1`,
+        /// inclusive.
+        ranges: Vec<(RReg, RReg)>,  // indexed by RegClass.
+    }
+        
+We can refer to a physical (real) register using the `RReg` type,
```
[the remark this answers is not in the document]
julian-seward1: The implementations use the term "real" throughout.  References to "physical" should be removed entirely, and a single comment made at the top, as you suggest, saying that "real" is really a reference to an "architected register".

docs/interface-proposal.md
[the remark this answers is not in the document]
julian-seward1: As far as I can see, the structure used here correctly describes all the ways a machine insn can interact with registers.  At least, for x86_64 and arm64 we have had no problem describing them in terms of the three sets.  A register which you describe as "clobbered" would be listed in the "defined" set.  Can you show me an insn or sequence thereof whose behaviour can't adequately be described in terms of these three sets?

julian-seward1: If a reg is in the "mod" set then it should not be in either of the "use" or "def" sets.  The regalloc implementation will tolerate them though; it removes "mod" from "def" and from "use" at startup.  Some examples might help clarify:

`(x86) add $1234, %v27`
Here we must say that `v27` is in the mod set, only.  Putting `v27` in both the def and use effectively tells the allocator that it can be mapped to different real registers on the input and output sides, which isn't true; and the allocator will assert when it comes to apply the mappings to the instruction. 

`(arm) add %v27, %v27, #1234`
Here we must list `v27` in the def set, because of its use as the destination, and also in the use set, because of its use as a source.  In this contrived example it would be also correct to instead put it, just once, in the mod set.  But that would be suboptimal, because it would force the allocator to map it to the same real register on both sides of the instruction, hence unnecessarily constraining the allocation.

As for "tied" .. I prefer not to use that phrase.  To my ear it suggests that an insn has two independent operands which, for whatever reason, need to be mapped to the the same real reg.  But that doesn't reflect reality: these (x86) insns don't have any such thing; they just have one operand which is both read and written (modified) by the insn.

docs/interface-proposal.md
julian-seward1: There is no provision for that.  At least for the bread-and-butter code generation we have so far been doing, there's been no need to support any such situation.  I am curious to know if there is a sequence that can't be described optimally using the def/mod/use scheme.  Do you know of any?

docs/interface-proposal.md
julian-seward1: I would prefer to change just to plain `regalloc`; to me it feels a bit strange to have a dot in a directory name.  (Dots belong to file names only!)

docs/interface-proposal.md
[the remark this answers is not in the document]
Amanieu: I can think of a few other cases where register allocation constraints can get complicated. It would be nice if the document could provide more specific guidance on how these should be handled.

- ARM: VFP register bank has overlapping S/D/Q registers.
- ARM: LDRD and STRD instructions allocate a consecutive pair of registers, which must start with an even numbered register.
- AArch64: The CASP (compare-and-swap pair) instruction introduced in ARMv8.1 also takes a consecutive pair of registers which must start with an even-numbered register.
- ARM: NEON structure load/store instruction take a sequence of 1 to 4 consecutive registers.
- AArch64: NEON structure load/store also take a sequence 1 to 4 consecutive registers, but the sequence is additionally allowed to wrap around (e.g. `LD4 {V30, V31, V0, V1}, [X0]`).

docs/interface-proposal.md
Amanieu: You can simulate an early-clobber by using a "modified" whose input comes from an `undef` pseudo-instruction.

docs/interface-proposal.md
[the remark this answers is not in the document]
julian-seward1: Ah, that's a good point re the Arm loads/stores and others.  Well, yes .. maybe we'll have to add elaboration to the interface to deal with such constraints, once the lack of them starts to become a limiting factor.

docs/interface-proposal.md
julian-seward1: I'm still curious to see an example of an early-clobber sequence that we'd want to present to the register allocator "as if" it were wrapped up into a single (pseudo-) machine instruction.  Maybe in some future support for inline assembly?

julian-seward1: Interesting.  Thanks for the examples.  I think for both examples, we could tell the allocator that the sequence modifies (both reads and writes) all of `out`, `in` and `addr`.  That would force it to put all 3 in different regs, which might be marginally suboptimal, but would certainly be correct.

PR #2, opened 2020-01-15
Integrate generic regalloc interface, and refactor backtracking regalloc and test-ISA to interact using it.

cfallin:
This PR (i) adds the interface defined in the design doc we merged in #1, (ii) moves the test-ISA types (`Inst`, `Block`, and `Func`) to the `tests` submodule, (iii) implements the `Function` trait on `Func`, thus serving as the first client of the generic API, and (iv) modifies the backtracking regalloc to operate through the trait instead of directly on the test ISA.

The linear-scan regalloc is still TODO, and there are probably some opportunities for cleanup, but I tried to make the translation as straightforward as I could. In the course of doing so, I played with the ergonomics a bit on the index/range type stuff; happy to iterate on that if it's not to taste.

@julian-seward1 or @bnjbvr, would you mind reviewing? Thanks!

### DIFF
--- a/src/backtracking.rs
+++ b/src/backtracking.rs
@@ -1126,6 +1126,6 @@ pub fn alloc_main<F: Function>(
     insns,
     target_map,
     clobbered_registers,
-    num_spill_slots: nextSpillSlot,
+    num_spill_slots: nextSpillSlot.get(),
   })
 }
--- a/src/backtracking.rs
+++ b/src/backtracking.rs
@@ -380,7 +380,7 @@ fn show_commit_tab(commit_tab: &Vec::<SortedRangeFragIxs>,
 */
 
 // Allocator top level.  This function returns a result struct that contains the final sequence of
-// instructions, possibly with fills/spills/moves spliced in and reundant moves elided, and with
+// instructions, possibly with fills/spills/moves spliced in and redundant moves elided, and with
 // all virtual registers replaced with real registers. Allocation can fail if there are
 // insufficient registers to even generate spill/reload code, or if the function appears to have
 // any undefined VirtualReg/RealReg uses.
@@ -1025,7 +1025,7 @@ pub fn alloc_main<F: Function>(
         debug_assert!(vlr_frag.first.pt.isReload());
         debug_assert!(vlr_frag.last.pt.isUse());
         debug_assert!(vlr_frag.first.iix == vlr_frag.last.iix);
-        let insnR = func.gen_reload(eli.slot, rreg);
+        let insnR = func.gen_reload(rreg, eli.slot);
         let whereToR = vlr_frag.first;
         spillsAndReloads.push((whereToR, insnR));
       }
@@ -1033,9 +1033,9 @@ pub fn alloc_main<F: Function>(
         debug_assert!(vlr_frag.first.pt.isReload());
         debug_assert!(vlr_frag.last.pt.isSpill());
         debug_assert!(vlr_frag.first.iix == vlr_frag.last.iix);
-        let insnR = func.gen_reload(eli.slot, rreg);
+        let insnR = func.gen_reload(rreg, eli.slot);
         let whereToR = vlr_frag.first;
-        let insnS = func.gen_spill(rreg, eli.slot);
+        let insnS = func.gen_spill(eli.slot, rreg);
         let whereToS = vlr_frag.last;
         spillsAndReloads.push((whereToR, insnR));
         spillsAndReloads.push((whereToS, insnS));
@@ -1044,7 +1044,7 @@ pub fn alloc_main<F: Function>(
         debug_assert!(vlr_frag.first.pt.isDef());
         debug_assert!(vlr_frag.last.pt.isSpill());
         debug_assert!(vlr_frag.first.iix == vlr_frag.last.iix);
-        let insnS = func.gen_spill(rreg, eli.slot);
+        let insnS = func.gen_spill(eli.slot, rreg);
         let whereToS = vlr_frag.last;
         spillsAndReloads.push((whereToS, insnS));
       }
@@ -1069,7 +1069,7 @@ pub fn alloc_main<F: Function>(
   let mut curB = mkBlockIx(0); // cursor in Func::blocks
 
   let mut insns: Vec<F::Inst> = vec![];
-  let mut target_map: Map<BlockIx, InstIx> = Map::default();
+  let mut target_map: TypedIxVec<BlockIx, InstIx> = TypedIxVec::new();
   let mut clobbered_registers: Set<RealReg> = Set::empty();
 
   for iix in func.insn_indices() {
@@ -1077,7 +1077,8 @@ pub fn alloc_main<F: Function>(
     // starting a new block?
     debug_assert!(curB.get() < func.blocks().len() as u32);
     if func.block_insns(curB).start() == iix {
-      target_map.insert(curB, mkInstIx(insns.len() as u32));
+      assert!(curB.get() == target_map.len());
+      target_map.push(mkInstIx(insns.len() as u32));
     }
 
     // Copy reloads for this insn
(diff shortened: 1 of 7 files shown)


src/interface.rs
```diff
@@ -0,0 +1,151 @@
...
+
+/// The result of register allocation.  Note that allocation can fail!
+pub struct RegAllocResult<F: Function> {
+  /// A new sequence of instructions with all register slots filled with real registers, and
+  /// spills/fills/moves possibly inserted (and unneeded moves elided).
+  pub insns: Vec<F::Inst>,
+
+  /// Basic-block start indices for the new instruction list, indexed by the original basic block
+  /// indices. May be used by the client to, e.g., remap branch targets appropriately.
+  pub target_map: Map<BlockIx, InstIx>,
```
julian-seward1: I had hoped that this could be a `TypedIxVec<BlockIx, InstIx>` since it is dense and zero-based and so doesn't need the generality of a `Map`.  Can it be changed?  I realised after emailing the draft to you that I'd made a typo here and confusingly written `TypedIxMap`.

src/data_structures.rs
```diff
@@ -204,14 +211,15 @@ pub struct TypedIxVec<TyIx, Ty> {
 impl<TyIx, Ty> TypedIxVec<TyIx, Ty>
 where
   Ty: Clone,
+  TyIx: From<u32>,
```
julian-seward1: Does this weaken the type-safety of `TypedIxVec`?  In particular, does it mean it can be indexed by anything that can be converted from a u32?  (I am unsure).  The design aim here was to ensure that the only allowable index type is `TyIx` and I'd really like to preserve that property.  Note that some of the more lame bits of ergonomics, in particular having to write `for ix in start .dotdot( end ) { .. }` will go away when once `trait Step` is available in stable Rust.

src/tests.rs
```diff
@@ -271,6 +1106,161 @@ impl Blockifier {
...
+  ) {
+    insn.mapRegs_D_U(
+      /* define-map = */ post_map, /* use-map = */ pre_map,
+    );
+  }
+
+  /// Allow the regalloc to query whether this is a move.
+  fn is_move(&self, insn: &Self::Inst) -> Option<(Reg, Reg)> {
+    match insn {
+      &Inst::Copy { dst, src } => Some((src, dst)),
```
julian-seward1: (nit) Ah, we should take care regarding the direction of assignment here (didn't occur to me when discussing the spec).  I have tried in general to have destinations "on the left" throughout, so I'd prefer if that was `... => Some((dst, src))`.

src/tests.rs
```diff
@@ -271,6 +1106,161 @@ impl Blockifier {
...
+      _ => None,
+    }
+  }
+
+  /// How many logical spill slots does the given regclass require?  E.g., on a
+  /// 64-bit machine, spill slots may nominally be 64-bit words, but a 128-bit
+  /// vector value will require two slots.  The regalloc will always align on
+  /// this size.
+  fn get_spillslot_size(&self, regclass: RegClass) -> SpillSlot {
```
julian-seward1: (really another q about the interface spec, not the implementation) .. if this function is returning a "number of somethings", shouldn't it return a `usize`, not `SpillSlot`?  At least in the demo implementation, `SpillSlot` is intended to denote the *identity* of a spill slot.  Does the interface spec intend a different meaning for it?

src/tests.rs
```diff
@@ -1116,15 +1116,21 @@ impl interface::Function for Func {
     &self.insns.elems()
   }
 
+  fn get_insn(&self, iix: InstIx) -> &Inst {
+    &self.insns[iix]
+  }
+
+  fn entry_block(&self) -> BlockIx {
```
julian-seward1: Quick sanity check: are we OK with assuming that block[0] is the entry block, for all uses?  Or is there any value in passing in the entry block index separately?

src/backtracking.rs
```diff
@@ -379,15 +379,16 @@ fn show_commit_tab(commit_tab: &Vec::<SortedRangeFragIxs>,
 }
 */
 
-// Allocator top level.  |func| is modified so that, when this function
-// returns, it will contain no VirtualReg uses.  Allocation can fail if there
-// are insufficient registers to even generate spill/reload code, or if the
-// function appears to have any undefined VirtualReg/RealReg uses.
+// Allocator top level.  This function returns a result struct that contains the final sequence of
+// instructions, possibly with fills/spills/moves spliced in and reundant moves elided, and with
```
julian-seward1: "reundant"

src/backtracking.rs
```diff
@@ -1042,63 +1068,64 @@ pub fn alloc_main(
...
 
   debug_assert!(curSnR == spillsAndReloads.len());
-  debug_assert!(curB.get() == func.blocks.len());
-  debug_assert!(curB.get() == newBlocks.len());
-
-  func.insns = newInsts;
-  func.blocks = newBlocks;
+  debug_assert!(curB.get() == func.blocks().len() as u32);
+
+  // Compute clobbered registers with one final, quick pass. Assert that all registers are real,
```
julian-seward1: This looks OK for an initial solution.  But it costs O(#insns) (= yet another pass over the whole function).  I'm pretty sure the same info can be generated in O(#real regs in the universe) by inspecting a table that the backtracking allocator creates (in short, its per-real-reg commitment table).  So I might rewrite this later.  I am also pretty sure that Benjamin can do the same for LSRA, by keeping track of which real regs have ever had a live range assigned to them.

julian-seward1: Also, the sanity-check component is sure to happen at emission time, since we will have to call get_hw_encoding() on all registers, and that will panic if given a virtual reg.

src/data_structures.rs
```diff
@@ -429,6 +432,14 @@ impl RegClass {
       _ => panic!("rc_from_u32"),
     }
   }
+
+  pub fn short_name(self) -> &'static str {
```
julian-seward1: Only of the most minor relevance, but .. I'd prefer to have different names for all classes, so as to avoid ambiguities reading virtual-registerised code printouts.  Maybe I, J, F, G, V?  Or I, II, F, FF, V?

julian-seward1:
That was quick!  On the whole these look fine to me, modulo a couple of minor nits.  I do notice there are formatting issues with some comments, but I can fix those up post-merge.

src/data_structures.rs
cfallin: Yes, it's perhaps a wider hole than we want -- replaced with a `Zero` trait so that the vec can build the range object from start (using `Zero`) to start+len (using `PlusN`).

src/tests.rs
cfallin: Fixed -- reversed the `is_move` return order and the args to `gen_spill` / `gen_reload` / `gen_move`.

src/tests.rs
cfallin: Yes, you're right. I suppose if we wanted to be really precise we'd have a separate `SpillSlotCount` type that `SpillSlot`'s `round_up`/`inc` can take, but that seems unnecessary to me here.

src/tests.rs
cfallin: It seems OK to wire this to 0 for the simple test ISA, unless you have more complex tests in mind in the future? The interface itself should allow weird entry points from the client.

src/backtracking.rs
cfallin: Sounds good, yes, I wrote this as I wasn't familiar enough with the internal data structures but please do replace it with a more direct derivation if you'd like!

I've added a TODO for that and I've removed the separate sanity-check-is-real loop over `used` regs.

src/interface.rs
```diff
@@ -0,0 +1,173 @@
...
+
+  /// Which real registers were overwritten? This will contain all real regs that appear as defs or
+  /// modifies in register slots of the output instruction list.
+  pub clobbered_registers: Set<RealReg>,
+
+  /// How many spill slots were used?
+  pub num_spill_slots: SpillSlot,
```
bjorn3: Should this be an `u32` too?

src/tests.rs
```diff
@@ -1,14 +1,867 @@
...
+use crate::interface;
+
 use crate::data_structures::{
-  i_add, i_addm, i_and, i_cmp_eq, i_cmp_ge, i_cmp_gt, i_cmp_le, i_cmp_lt,
-  i_copy, i_finish, i_goto, i_goto_ctf, i_imm, i_load, i_mod, i_mul, i_print_i,
-  i_print_s, i_shr, i_store, i_sub, i_subm, is_goto_insn, mkInstIx, mkRealReg,
-  mkTextLabel, mkVirtualReg, remapControlFlowTarget, BinOp, Func, Inst, InstIx,
-  Label, RealReg, Reg, RegClass, SpillSlot, TypedIxVec, AM, AM_R, AM_RI, AM_RR,
-  RI, RI_I, RI_R,
+  mkBlockIx, mkInstIx, mkRealReg, mkSpillSlot, mkVirtualReg, BlockIx, InstIx,
```
bjorn3: The standard style is 4 spaces as indentation.

src/interface.rs
cfallin: Fixed, thanks!

src/tests.rs
cfallin: I ran everything through `rustfmt` (via `cargo fmt`); it seems that the `rustfmt.toml` file which was added recently specifies 2-space indentation. Perhaps we should change that if the project standard is still 4 spaces?

cfallin:
Will go ahead and merge as per @julian-seward1. Thanks!

PR #7, opened 2020-01-15
Exposed crate as a library, with test-runner in separate binary.

cfallin:
Also made a few traits public to avoid errors when using API from another crate.

This change allows the crate to be pulled into a Cranelift build as a
dependency.


cfallin:
Created PR as a record, but merging right away as per Julian (loose reviewing requirements at the moment).

PR #20, opened 2020-02-03
Add function liveins/liveouts to interface and backtracking allocator.

cfallin:
This change allows the Function to specify that certain registers
(probably real registers, though the interface is general) are live on
function entry -- i.e., have an implicit def before the function runs --
and/or are live on function exit -- i.e., have an implicit use after the
function returns. (Every block without successors is considered a return
point.)

This allows code generators that expect function arguments in certain
real registers, and return the function return value(s) in certain real
registers, to properly use the args and define the retvals without the
allocator complaining (undefined args) or miscompiling (spilled
retvals).

This change modifies the interface and updates the backtracking
allocator, but not the linear-scan allocator. The builtin tests do not
use this feature; I plan to test it from within Cranelift for now
(though we should add better tests here that exercise some sort of ABI,
too).

### DIFF
--- a/bin/test_framework.rs
+++ b/bin/test_framework.rs
@@ -1617,6 +1617,13 @@ impl regalloc::Function for Func {
     self.insns[last_insn].getTargets()
   }
 
+  fn is_ret(&self, insn: InstIx) -> bool {
+    match &self.insns[insn] {
+      &Inst::Finish { .. } => true,
+      _ => false,
+    }
+  }
+
   /// Provide the defined, used, and modified registers for an instruction.
   fn get_regs(&self, insn: &Self::Inst) -> regalloc::InstRegUses {
     let (d, m, u) = insn.get_reg_usage();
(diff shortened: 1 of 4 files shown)


cfallin:
@julian-seward1, would you mind reviewing? Thanks!

cfallin:
Updated based on comments on Matrix -- thanks!

bjorn3:
What is the Matrix channel? Or is it only for Mozilla employees?

cfallin:
@bjorn3: Mozilla's Matrix is open to everyone now (see https://chat.mozilla.org/) and is our replacement for IRC. What I referred to here was that Julian gave some feedback on the patch in an individual chat over Matrix, though, and AFAIK there's no CraneLift channel there.

bjorn3:
> Mozilla's Matrix is open to everyone now (see https://chat.mozilla.org/) and is our replacement for IRC.

I searched for it, but couldn't find it.

> AFAIK there's no CraneLift channel there.

:+1:

PR #30, opened 2020-03-11
Register allocator checker

cfallin:
This PR adds a register allocator checker: an analysis that does a dataflow analysis of the register allocator's output, including spills, reloads, moves, and mapped real registers, to ensure that the resulting program has the same semantics as the virtual-register-based input.

The checker is wired into the backtracking allocator (via the `BacktrackingChecked` alternative algorithm choice, because we don't have a separate way to specify flags). Integration into linear-scan should be pretty straightforward too; I just haven't done it yet because the top-level is a bit different (doesn't use `edit_inst_stream`).

### DIFF
--- a/bin/minira.rs
+++ b/bin/minira.rs
@@ -120,7 +120,7 @@ mod test_cases;
 mod test_framework;
 mod validator;
 
-use regalloc::{allocate_registers, RegAllocAlgorithm};
+use regalloc::{allocate_registers, RegAllocAlgorithm, RegAllocError};
 use test_framework::{make_universe, run_func, RunStage};
 use validator::check_results;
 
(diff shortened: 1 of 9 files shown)


julian-seward1:
This looks awesome (and very quickly put together!).  I look forward to trying it on our current non-working inputs.  I have no particular comments on the low level details/coding, and would be in favour of landing it.  I do have three general points though:

* I see places where instructions created by `func.gen_{spill,reload,move}` are now replaced by an `InstToInsert`, which looks like some abstractified version of s/r/m insns that presumably is later transformed into a target-specific version somehow.  (Right?)  The concern is whether this might slow down the allocator in as much as it constitutes extra copying of information.  But I think I can live with it .. if it is an inefficiency, there are much bigger ones elsewhere in the allocator as a whole, that first need to be dealt with.  Plus, in most cases, there won't be much, if any, spilling.

* The current BT algorith, and I suspect LSRA too, only uses each spill slot for one virtual live range.  It never recycles them.  For a production implementation that's obviously unacceptable -- I plan to add a simple spill slot occupancy tracker/allocator at some point.  Q: does your analysis actually depend on the current each-spilled-VLR-has-its-own-spill-slot behaviour?  It would be a shame if so, since that'll mean it will break when that is fixed.

* I see you added new `RegAllocAlgorithm` allocator kinds.  I have been contemplating a more general scheme, at least for this early stage development, where we provide a general and easy to use way to enable cross-checking functionality.  I say this because the current sources already provide a cross-check mechanism for BT's core CommitmentMap machinery, and I am about to land a second mechanism for cross-checking of dominator-set computation.  So your checker could be a third case for this.

   The scheme I have been contemplating simply involves an env var that is set at run time:
```
  REGALLOC_CHECK=   some subset of {btcm,dom,overall,timing}
  where 
        btcm     is BT's CommitmentMap
        dom      is dominator-set computation
        overall  is this analysis
        timing   is kind of anomalous, but means: print summaries of RA pass
                 timings without all the debug spew.  I think this will be
                 important.
```
(this should be indented; oh well)
Then we could use cross checking in whatever embedding (test framework, wasmtime, SpiderMonkey) uniformly and without the hassle of having to try and wire up control settings separately in each embedding.  For this reason I'm not entirely enthusiastic about the change to `RegAllocAlgorithm`.

bnjbvr:
I haven't looked at the code itself, but great news this is working!

Regarding selection of features: env variables have their limits, and it would be nice to allow users of this crate to enable the supplementary checks, if needed. So we could have a real way to give configuration options, and in the minira binary, these could be enabled through env variables and/or binary flags. If that makes sense, I'm happy to implement this.

cfallin:
Just rebased on top of latest linear-scan changes. Thanks for feedback! A few answers:

> I see places where instructions created by func.gen_{spill,reload,move} are now replaced by an InstToInsert, which looks like some abstractified version of s/r/m insns that presumably is later transformed into a target-specific version somehow. (Right?)

Indeed, and hopefully this isn't too much overhead -- in fact the list of ops to insert will likely be more compact in this form than as the full `arm64::Inst` struct. In any case the alternative was to add more API surface to `Inst` to query whether something is a spill/reload; the current approach feels cleaner.

>  Q: does your analysis actually depend on the current each-spilled-VLR-has-its-own-spill-slot behaviour? It would be a shame if so, since that'll mean it will break when that is fixed.

No, it should be fully flexible w.r.t. spill-slot sharing. The content of each spill-slot is tracked at each program-point (i.e., it's a flow-sensitive analysis).

Regarding configuration options: I generally agree with Ben that it'd be nice to have a central configuration infrastructure, with different ways of setting flags. I wouldn't necessarily want to insert surprising dependencies on environment variables into a program that uses this library, at least not without explicit opt-in to that.

If you both don't mind, though, perhaps we could land this as-is first, and then work out the config stuff as a separate issue? Mostly I'm motivated by wanting to get this going in the Cranelift branch without local tweaks to build with this PR :-) Alternatively, if we don't want to clutter the algorithm-choice enum, we could add an environment-variable check for now.

julian-seward1:
Fine by me to land as-is.

cfallin:
Sounds good! I'll create an issue to track the configuration-infra / flags / env vars question.

PR #35, opened 2020-03-23
Add interface to allow regalloc to query proper slot number for combined slot

cfallin:
As per discussion with @julian-seward1 on Zulip, this allows the register allocator to query the client: when it needs to use an N-slot space (e.g., two slots, for a 128-bit vector reg), it needs to know how to refer to this combined slot in spill and reload instructions. Depending on the underlying address computation / stack frame layout that the client uses, this may be `base` (first constituent slot), `base + len - 1` (last constituent slot), or maybe even something else.

### DIFF
--- a/bin/test_framework.rs
+++ b/bin/test_framework.rs
@@ -1972,6 +1972,10 @@ impl regalloc::Function for Func {
     None
   }
 
+  fn combined_slot(first_slot: SpillSlot, _num_slots: u32) -> SpillSlot {
+      first_slot
+  }
+
   fn func_liveins(&self) -> Set<RealReg> {
     Set::empty()
   }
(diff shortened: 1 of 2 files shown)


julian-seward1:
Thanks!
I had just been considering something even simpler.  Could the client simply pass (maybe on request) to the RA, one of the following two values:
`enum SpillAreaOrientation { SpillAreaIndexedFromTop, SpillAreaIndexFromBottom }`
(where the current implied semantics is `SpillAreaIndexedFromTop`)

and we specify in the interface:
for the case `SpillAreaIndexedFromBottom`, the RA can assume that `&spill_slot[0]` is 16-aligned
and 
for the case `SpillAreaIndexedFromTop`, the RA can assume that `&spill_slot[0] + sizeof(spill_slot[0])` is 16-aligned
and then let the RA take care of creating the correct numbers for spill/reload insns
(it's hardly a big deal).

cfallin:
Closing outdated PR.

PR #39, opened 2020-04-10
Minor external interface tidying

sunfishcode:
This updates a comment, and tweaks the public interface to just export the contents of the `interface` rather than also `interface` itself, to make the front page of https://docs.rs/regalloc/ more informative.


bnjbvr:
Thanks!

PR #41, opened 2020-04-10
Merge the interface module into lib.rs.

sunfishcode:
lib.rs is already the mechanism to describe the public interface of a
crate, so use it directly.


cfallin:
Looks good; thanks!

PR #48, opened 2020-04-14
Fixes #43: Use markdown code comments for code;

### LINKED ISSUE #43: Use markdown-style code formatting in comments.
The current code uses '|' to indicate code formatting within comments, for example [here](https://github.com/bytecodealliance/regalloc.rs/blob/master/lib/src/analysis.rs#L43).

In Rust code, including Cranelift, and Rust code in Firefox such as WebRender, Servo, and Cranelift JIT support, it is conventional to use backticks, ``` ` ``` for this purpose. In addition to being the convention, this would allow such comments to be formatted appropriately in documentation, for example here: https://docs.rs/regalloc/0.0.17/regalloc/interface/enum.AnalysisError.html#variant.IllegalRealReg.

bnjbvr:
This was done using the following sed line:

    sed -r "s%^([^\|]*//[^\|]*)\|(.*)\|(.*)%\1\`\2\`\3%"

applied repeatitively until it reaches a fixed point, on each rust file.

### DIFF
--- a/lib/src/inst_stream.rs
+++ b/lib/src/inst_stream.rs
@@ -176,7 +176,7 @@ fn map_vregs_to_rregs<F: Function>(
             return true;
         }
         // A spill-related ("bridge") frag.  There are three possibilities,
-        // and they correspond exactly to |BridgeKind|.
+        // and they correspond exactly to `BridgeKind`.
         if frag.first.pt.is_reload() && frag.last.pt.is_use() && frag.last.iix == frag.first.iix {
             // BridgeKind::RtoU
             return true;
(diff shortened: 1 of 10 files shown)


lib/src/inst_stream.rs
```diff
@@ -162,7 +162,7 @@ fn map_vregs_to_rregs<F: Function>(
 
     //fn showMap(m: &Map<VirtualReg, RealReg>) -> String {
     //  let mut vec: Vec<(&VirtualReg, &RealReg)> = m.iter().collect();
-    //  vec.sort_by(|p1, p2| p1.0.partial_cmp(p2.0).unwrap());
+    //  vec.sort_by(`p1, p2` p1.0.partial_cmp(p2.0).unwrap());
```
sunfishcode, line 165: This one should stay as `|`.

PR #50, opened 2020-04-21
Add `orig_insn_map`: mapping from original to new insn indices.

cfallin:
This is needed in the Cranelift client code in order to track
source-location mapping for debug info: we need to be able to shuffle
the source locations into the new locations after the regalloc has done
its instruction-stream editing. The `target_map` result is not quite
good enough, because it only provides old --> new mappings at a basic
block granularity.

### DIFF
--- a/lib/src/data_structures.rs
+++ b/lib/src/data_structures.rs
@@ -435,21 +435,38 @@ macro_rules! generate_boilerplate {
             #[allow(dead_code)]
             #[inline(always)]
             pub fn new(n: u32) -> Self {
+                debug_assert!(n != u32::max_value());
                 Self::$TypeIx(n)
             }
             #[allow(dead_code)]
             #[inline(always)]
-            pub fn max_value() -> Self {
-                Self::$TypeIx(u32::max_value())
+            pub const fn max_value() -> Self {
+                Self::$TypeIx(u32::max_value() - 1)
             }
             #[allow(dead_code)]
             #[inline(always)]
-            pub fn min_value() -> Self {
+            pub const fn min_value() -> Self {
                 Self::$TypeIx(u32::min_value())
             }
             #[allow(dead_code)]
             #[inline(always)]
+            pub const fn invalid_value() -> Self {
+                Self::$TypeIx(u32::max_value())
+            }
+            #[allow(dead_code)]
+            #[inline(always)]
+            pub fn is_valid(self) -> bool {
+                self != Self::invalid_value()
+            }
+            #[allow(dead_code)]
+            #[inline(always)]
+            pub fn is_invalid(self) -> bool {
+                self == Self::invalid_value()
+            }
+            #[allow(dead_code)]
+            #[inline(always)]
             pub fn get(self) -> u32 {
+                debug_assert!(self.is_valid());
                 match self {
                     $TypeIx::$TypeIx(n) => n,
                 }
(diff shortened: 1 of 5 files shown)


julian-seward1:
I see why this would be wanted.  But it's another 8 bytes per instruction that the allocator needs to sit on and then drag through the machine's caches, and we're already way slow.  Is this mapping needed in production runs?  If not, I'd prefer to have it implemented similar to the existing `request_block_annotations` facility -- that is, the client passes in a flag requesting the mapping, and we return it as an `Option<TypedIxVec<..>>` instead.  That way we can have our cake and eat it, albeit at a little more implementation complexity.

lib/src/inst_stream.rs
```diff
@@ -463,15 +471,18 @@ fn add_spills_reloads_and_moves<F: Function>(
             && insts_to_add[curITA].point == InstPoint::new_reload(iix)
         {
             insns.push(insts_to_add[curITA].inst.construct(func));
+            orig_insn_map.push(None);
```
bnjbvr, line 474: I suppose you maintain a mapping of new instructions to old instructions, thus needing to push None; could it also work with the inverted mapping (old instructions to new instructions), removing the need for None elements in the array?

lib/src/inst_stream.rs
```diff
@@ -447,6 +454,7 @@ fn add_spills_reloads_and_moves<F: Function>(
 
     let mut insns: Vec<F::Inst> = vec![];
     let mut target_map: TypedIxVec<BlockIx, InstIx> = TypedIxVec::new();
+    let mut orig_insn_map: TypedIxVec<InstIx, Option<InstIx>> = TypedIxVec::new();
```
bnjbvr, line 457: Could you pre-reserve the full array here to `insts_to_add.len()`?

lib/src/inst_stream.rs
bjorn3: That would make debuginfo generation slower, as you would need to iterate through the whole map to find the original instruction corresponding to the instruction you are currently emitting lineinfo for. Also you may want to attribute those extra insts to `iix`.

bnjbvr: Good point about attributing those extra insts to `iix`.

However i think there's still a linear way to find the source information for a given new inst, since the source locations are handled all at once: start with the first old instruction, remember what its new instruction is, then on the second instruction, you see the new inst mapping to the second instruction; you then know that all the instructions between first's new instruction and second's previous new instruction all map to the first old instruction.

cfallin:
@julian-seward1 re: only doing this when necessary: it appears that the `SourceLoc` info is always needed, as shown for example by the panic caused by code we had to disable to get wasmtime to run at all. It's possible we could push the notion of "source locations are optional" further up the stack, but so far our clients seem to be built assuming it's always there.

Per @bnjbvr in the related Cranelift PR, we have `SourceLoc::default()` so it's actually just 4 bytes per instruction -- not ideal, but IMHO not the end of the world either, relative to our core algorithms' costs...

cfallin:
Err, I misspoke w.r.t. the default val (we're in the regalloc crate here -- we're operating with `InstIx`'s); perhaps we could have a "no instruction" special value in the typed index definitions to allow for this savings?

julian-seward1:
@cfallin ah well, so be it.  If we have to have it, we have to have it.  I'll try and quantify the cost in the next round of RA profiling.

cfallin:
Just added a `::not_present()` option to index types (turns out there were several ad-hoc 0xFFFFFFFF not-present literals already); PTAL!

lib/src/inst_stream.rs
cfallin: Done!

cfallin: (Actually to the original plus to-add, i.e. `func.insn_indices().len() + insts_to_add.len()`)

lib/src/data_structures.rs
```diff
@@ -435,61 +435,83 @@ macro_rules! generate_boilerplate {
...
             pub fn dotdot(&self, last_plus1: $TypeIx) -> Range<$TypeIx> {
+                debug_assert!(self.is_present());
                 let len = (last_plus1.get() - self.get()) as usize;
                 Range::new(*self, len)
             }
         }
         impl fmt::Debug for $TypeIx {
             fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-                write!(fmt, "{}{}", $PrintingPrefix, &self.get())
+                if *self == Self::not_present() {
```
bnjbvr, line 490: nit: `if !self.is_present()`

julian-seward1: Ah, I should have spoken earlier.  I would have preferred `invalid` as a name.

bnjbvr: Agreed, `invalid` is nice too!

cfallin: Agreed, that's better -- changed to `is_valid()` / `is_invalid()`, and the static constructor to `invalid_value()` to match `min_value()` / `max_value()`'s naming scheme.

PR #54, opened 2020-04-27
Add options to configure the register allocator

bnjbvr:
This is an API breaking change. This allows selecting either a
preconfigured register allocator with sensible defaults, or to configure
options common to all the allocators and specific to each allocator.

### DIFF
--- a/bin/minira.rs
+++ b/bin/minira.rs
@@ -4,12 +4,12 @@ mod test_cases;
 mod test_framework;
 mod validator;
 
-use regalloc::{allocate_registers, RegAllocAlgorithm};
+use regalloc::{allocate_registers_with_opts, Algorithm, BacktrackingOptions, Options};
 use test_framework::{make_universe, run_func, RunStage};
 use validator::check_results;
 
 use clap;
-use log::{self, error, info};
+use log::{self, error};
 use pretty_env_logger;
 
 //=============================================================================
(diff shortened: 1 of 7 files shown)


lib/src/linear_scan.rs
```diff
@@ -2249,6 +2265,7 @@ pub fn run<F: Function>(
     func: &mut F,
     reg_universe: &RealRegUniverse,
     use_checker: bool,
+    _opts: &LinearScanOptions,
```
bjorn3, line 2268: Maybe

```suggestion
    &LinearScanOptions: &LinearScanOptions,
```

:)

lib/src/linear_scan.rs
```diff
@@ -50,6 +51,21 @@ where
...
 }
 
+#[derive(Clone)]
+pub struct LinearScanOptions {}
+
+impl default::Default for LinearScanOptions {
+    fn default() -> Self {
+        Self {}
+    }
+}
```
bjorn3, line 61:
```suggestion
pub struct LinearScanOptions;

impl default::Default for LinearScanOptions {
    fn default() -> Self {
        Self
    }
}
```

lib/src/linear_scan.rs
bnjbvr: wow, what is this syntax!

bjorn3: Yes, parameters can be any irrefutable pattern. For example:

```rust
fn a((a, (b, ())): (u8, (u8, ()))) {}
```

Amanieu:
Consider using `#[non_exhaustive]` if you want to keep open the possibility of adding options in the future.

bnjbvr:
> Consider using #[non_exhaustive] if you want to keep open the possibility of adding options in the future.

TIL, thanks! This removes the ability to use function update syntax with a default value, though, so I'm then wondering how users can construct the options. The builder pattern would work, yet it is a bit heavyweight for this use case.

bjorn3:
Is `#[non_exhaustive]` really necessary? Both major and minor version are 0, which means that cargo will consider any increment of the patch version a breaking change anyway. If the minor version becomes nonzero, I expect that new features get a minor bump, which is considered a breaking change when the major version is 0.

bnjbvr:
It's a good question. Actually I now think that the current approach with `Default` is better, because:

- people who don't care about some new flags can use the `Default::default()` with a functional update (`{ something: true, ..Default::default() }`), so as to set only the options they're interested in, and get the default for the rest, without any change to their code.
- people who want to be in full control can not use the `Default` impl, and from one version bump to the other, consider what they want for the new configuration options.

lib/src/linear_scan.rs
bnjbvr: I plan to make use of the lsra options soon, so I'll leave it as is in the meanwhile.

bin/minira.rs
```diff
@@ -73,22 +73,23 @@ fn main() {
         }
     };
 
-    let reg_alloc_kind = match matches.value_of("algorithm").unwrap() {
-        "bt" => {
+    let algorithm = matches.value_of("algorithm").unwrap();
+    let opts = match algorithm {
+        "bt" | "btc" => {
             info!("Using the backtracking allocator");
```
cfallin, line 79: Would it be better to print the whole `opts` struct to info/debug out instead? (This is just log output on our utility tool so not a huge deal either way; I just tend toward more information in logs usually)

julian-seward1: +1 for dumping the whole thing at `debug` but not `info`; `info` is most useful if it's relatively concise, and so far we do about a page of text per fn, which is nicely concise.

bnjbvr: I even removed it; the output in `allocate_with_opts` made this redundant.

PR #55, opened 2020-04-28
Use incrementally-updated, dense vreg-to-rreg maps.

cfallin:
This PR changes the interface of `map_regs()` to take, instead of two
Maps (a hashmap), two abstracted `RegVRMap` instances bundled together.
The implementation of this map is a dense vector indexed by vreg; in
almost all cases this should be faster than a hashmap.

Furthermore, this PR changes the instruction-stream editing to maintain
running incrementally-updated pre- and post-instruction maps, rather
than cloning the full hashmap twice for every instruction. A preliminary
investigation of Cranelift's new backend using DHAT shows that this
clone is very costly in terms of allocation; this interface and
algorithm change should significantly reduce the cost.

### DIFF
--- a/lib/src/data_structures.rs
+++ b/lib/src/data_structures.rs
@@ -9,6 +9,7 @@ use std::collections::VecDeque;
 use std::fmt;
 use std::hash::Hash;
 use std::marker::PhantomData;
+use std::mem;
 use std::ops::Index;
 use std::ops::IndexMut;
 use std::slice::{Iter, IterMut};
@@ -640,12 +641,14 @@ pub struct Reg {
     do_not_access_this_directly: u32,
 }
 
+static INVALID_REG: u32 = 0xffffffff;
+
 impl Reg {
     pub fn is_virtual(self) -> bool {
-        (self.do_not_access_this_directly & 0x8000_0000) != 0
+        self.is_valid() && (self.do_not_access_this_directly & 0x8000_0000) != 0
     }
     pub fn is_real(self) -> bool {
-        !self.is_virtual()
+        self.is_valid() && !self.is_virtual()
     }
     pub fn new_real(rc: RegClass, enc: u8, index: u8) -> Self {
         let n = (0 << 31) | (rc.rc_to_u32() << 28) | ((enc as u32) << 8) | ((index as u32) << 0);
@@ -662,10 +665,29 @@ impl Reg {
             do_not_access_this_directly: n,
         }
     }
+    pub fn invalid() -> Reg {
+        Reg {
+            do_not_access_this_directly: INVALID_REG,
+        }
+    }
+    pub fn is_invalid(self) -> bool {
+        self.do_not_access_this_directly == INVALID_REG
+    }
+    pub fn is_valid(self) -> bool {
+        !self.is_invalid()
+    }
+    pub fn is_virtual_or_invalid(self) -> bool {
+        self.is_virtual() || self.is_invalid()
+    }
+    pub fn is_real_or_invalid(self) -> bool {
+        self.is_real() || self.is_invalid()
+    }
     pub fn get_class(self) -> RegClass {
+        debug_assert!(self.is_valid());
         RegClass::rc_from_u32((self.do_not_access_this_directly >> 28) & 0x7)
     }
     pub fn get_index(self) -> usize {
+        debug_assert!(self.is_valid());
         // Return type is usize because typically we will want to use the
         // result for indexing into a Vec
         if self.is_virtual() {
@@ -675,6 +697,7 @@ impl Reg {
         }
     }
     pub fn get_hw_encoding(self) -> u8 {
+        debug_assert!(self.is_valid());
         if self.is_virtual() {
             panic!("Virtual register does not have a hardware encoding")
         } else {
@@ -682,14 +705,16 @@ impl Reg {
         }
     }
     pub fn as_virtual_reg(self) -> Option<VirtualReg> {
-        if self.is_virtual() {
+        // Allow invalid virtual regs as well.
+        if self.is_virtual_or_invalid() {
             Some(VirtualReg { reg: self })
         } else {
             None
         }
     }
     pub fn as_real_reg(self) -> Option<RealReg> {
-        if self.is_real() {
+        // Allow invalid real regs as well.
+        if self.is_real_or_invalid() {
             Some(RealReg { reg: self })
         } else {
             None
@@ -698,21 +723,27 @@ impl Reg {
     pub fn show_with_rru(self, univ: &RealRegUniverse) -> String {
         if self.is_real() && self.get_index() < univ.regs.len() {
             univ.regs[self.get_index()].1.clone()
-        } else {
+        } else if self.is_valid() {
             format!("{:?}", self)
+        } else {
+            "rINVALID".to_string()
         }
     }
 }
 
 impl fmt::Debug for Reg {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        write!(
-            fmt,
-            "{}{}{}",
-            if self.is_virtual() { "v" } else { "r" },
-            self.get_index(),
-            self.get_class().short_name(),
-        )
+        if self.is_valid() {
+            write!(
+                fmt,
+                "{}{}{}",
+                if self.is_virtual() { "v" } else { "r" },
+                self.get_index(),
+                self.get_class().short_name(),
+            )
+        } else {
+            write!(fmt, "rINVALID")
+        }
     }
 }
 
@@ -746,6 +777,24 @@ impl RealReg {
     pub fn to_reg(self) -> Reg {
         self.reg
     }
+    pub fn invalid() -> RealReg {
+        RealReg {
+            reg: Reg::invalid(),
+        }
+    }
+    pub fn is_valid(self) -> bool {
+        self.reg.is_valid()
+    }
+    pub fn is_invalid(self) -> bool {
+        self.reg.is_invalid()
+    }
+    pub fn maybe_valid(self) -> Option<RealReg> {
+        if self == RealReg::invalid() {
+            None
+        } else {
+            Some(self)
+        }
+    }
 }
 impl fmt::Debug for RealReg {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
@@ -776,6 +825,24 @@ impl VirtualReg {
     pub fn to_reg(self) -> Reg {
         self.reg
     }
+    pub fn invalid() -> VirtualReg {
+        VirtualReg {
+            reg: Reg::invalid(),
+        }
+    }
+    pub fn is_valid(self) -> bool {
+        self.reg.is_valid()
+    }
+    pub fn is_invalid(self) -> bool {
+        self.reg.is_invalid()
+    }
+    pub fn maybe_valid(self) -> Option<VirtualReg> {
+        if self == VirtualReg::invalid() {
+            None
+        } else {
+            Some(self)
+        }
+    }
 }
 impl fmt::Debug for VirtualReg {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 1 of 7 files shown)


lib/src/data_structures.rs
```diff
@@ -1085,6 +1081,130 @@ impl RegVecsAndBounds {
...
+/// consumed by the client's `map_reg()` function that replaces register
+/// references in each instruction according to this map after allocation is
+/// complete.
+///
+/// To avoid the need to hold an `Option<RealReg>` in every slot, for
+/// performance, we require a "default register" that we store in sparse
+/// gaps and return when no mapping is found. The register allocator should
+/// never call `map_regs()` on an instruction with a vreg reference not
+/// present in the map, so this default register should not matter.
+#[derive(Clone, Debug)]
```
bjorn3, line 1097: Does this need to be `Clone`?

julian-seward1:
Thank you first of all for working on this.

I was (am!) fascinated and pleased to see that you independently arrived at the same main-array-plus-deltas-that-are-periodically-flushed solution that I had also been thinking about.

The design I had in mind differs from your implementation in one aspect, though.  I had envisaged only having one main array and one deltas array.  The main array would be what the existing code calls `map`.  Then, at the point in the loop that `map` is cloned into `map_uses`, we leave the main vec unchanged and instead put changes into the delta array.  Once we get to the point where `map` is cloned into `map_defs`, we immediately apply the (main vec, deltas vec) pair to the instruction.  So, "use" mentions in the insn are looked up in the main vec only, whereas "def" mentions are looked up in the main vec as fronted by the delta vec.

Once done, we push the deltas in to the main vec, empty the delta vec, and move on.

So I had been hoping to get away with just one vec (== half the amount of memory and D1 misses).

Now .. that said .. I didn't try to implement my plan, so this might all be drive-by blah-blah nonsense.  I see in particular that there are complications to do with driving the checker, that I hadn't considered.  But let me ask you: are you sure it isn't possible to do all this with just one main vec and one delta vec?  Or one main vec and two delta vecs?  If so I think that would be preferable.

From a naming perspective, I had thought to name the client-visible mappings -- what you call `RegVRMaps` -- as `RegUsageMapper`.  This is so as to emphasise the commonality with `RegUsageCollector` that is presented via `get_regs`.

Also, I would prefer to retain the `defs` and `uses` naming rather than using `pre` and `post`.  `pre` and `post` as applied to an instruction don't trigger any immediate connection for me in the general parlance of instructions, whereas `def` and `use` does.  All of the allocator internals that deal with instruction-register interactions, including the `RegUsageMapper`, is phrased in terms of `def`, `use` (and `mod`).  Would you be happy to retain `defs` and `uses` here?

cfallin:
Interesting -- happy to hear you had been thinking along the same lines!

Regarding the question of two main arrays vs. one plus deltas: the thought that guided me to this design was that we should avoid anything more expensive than a direct array access for register lookups; I wasn't sure how efficient an overlay would be. (I assume that if we used the deltas as overlay, we would also keep a hashmap so that we wouldn't need to do a linear search through the deltas?) The other benefit of two full arrays is that the pre- and post- (use- and def-) cases in instruction impls take identical data structures and can share code. We could hide a conditional check inside `RegUseMapper` (née `RegVRMap`) to use an overlay if specified but that adds a branch to each lookup.

Point taken re: working-set size, though. These arrays are indexed by vreg so will become large as the function body does... but then so does everything else (main instruction array, fragments, ...) -- I suppose every little bit helps regardless.

Anyway, I'm happy to go with single large array + deltas array + hashmap overlay (indexing the deltas by vreg) if you'd prefer -- let me know what you think based on the above.

bjorn3:
If using a hashmap, you should probably use [`fnv::FnvHashMap`](https://docs.rs/fnv/1.0.6/fnv/). It is faster than the default hasher at the cost of more hash collisions for larger keys. As the keys are small, this shouldn't be a problem.

julian-seward1:
So -- with the usual proviso that I might misunderstand -- my belief is that the deltas array will almost always have either zero or one entry in it.  So it's real cheap to "search" and no hashmappery is needed.  Here's why:

The delta array only ever carries the differences in the main map between the "use" side of an instruction and the "def" side.  Specifically, the main map describes the use side, and the delta+map combo describes the def side.  So the number of entries in the delta array must be bounded by the number of virtual registers that any instruction can write.  Most insns write either zero or one registers; I'd be hard put to think of one that writes two -- maybe one of the arm64 load-and-move-base-register insns.  More than two, I cannot think of any.

The only insn that gives me pause for thought are calls.  We annotate these with all the real regs they (allegedly) write, which is a lot.  But those are all real regs, not virtual regs, so they don't appear in either the main map or the deltas array.

And so checking a len-1 deltas array for handling an insn's def'd register is going to be a handful of host insns at most: bounds check, load vreg from first pair, compare (== tag check), branch, load rreg from first pair.  Maybe 10 host insns?

Also, most insns read more regs than they write.  And we only pay the delta-array cost for the writes.

Now granted, it could be more complex if we're driving the checker.  But we could accept a bit of slowdown there; and in any case it can't get *really* bad since the deltas array will be flushed into the main map each time round the loop.

julian-seward1:
Hmm, now I worry my analysis isn't correct.  Maybe it's the number of live ranges that can start/end at an insn that determines the upper limit on delta array size?  And because of block boundaries, maybe many live ranges really will start/end at the block's boundary instructions?  Even if so, isn't #liveranges = O(#insns), so there's some amortisation effect?  I don't know.

cfallin:
Yes, the overlay size would scale with the number of live-range start/end events. I'll instrument things and take a quick look with a few tests (`regex-rs`, `bzip2`, and the built-in testcases at least); if it's small, then the "smallset" (linear array) approach makes a lot of sense, I agree.

cfallin:
Here's a histogram of overlay size (the delta from use to def maps for `regex-rs.wasm` from my tests repo:

```
  28338 0
  86977 1
 111442 2
  42272 3
  23081 4
   3966 5
   3338 6
   2588 7
   2335 8
   1907 9
   1635 10
   1243 11
   1292 12
   1003 13
    897 14
    729 15
    687 16
    536 17
    404 18
    330 19
    274 20
    224 21
    222 22
    230 23
    185 24
    136 25
    166 26
    101 27
    121 28
    119 29
    101 30
     79 31
     78 32
     48 33
     52 34
     28 35
     37 36
     22 37
     14 38
     16 39
     24 40
     22 41
     17 42
     18 43
     14 44
     10 45
      5 46
      9 47
      4 48
     11 49
      2 50
      1 51
```

(added a `println!()` to `apply_post_to_pre()`, grep'd/awk'd out the number, then `sort -n | uniq -c`)

So it seems that the majority of deltas are small (the median is 2) but we do have a worst-case tail we need to worry about; maybe a good middle ground is to sort the deltas by vreg if over a threshold and do a binary search (no need for extra mode-bit because it's implicit in length if threshold is fixed). I'll hack this up...

cfallin:
(For the record, `bz2`'s data looked very similar; max of 51 (is that the total number of registers, I bet?), median of 2, long tail.)

lib/src/data_structures.rs
cfallin: Nope, thanks, removed (was just out of habit; I tend to make most things `Clone` and `Debug` by default, but there's no reason to clone this much data).

cfallin:
Updated -- PTAL! I took a hybrid approach, sorting and binary-searching only when the delta is large enough. The logic is a little subtle because of multiple updates, and the need for the last update to win. (E.g.: set to tombstone on delete when live range for a vreg ends, then set to another reg when a new live range for that vreg begins. I don't know if this can happen according to the higher-level regalloc logic, but the delta data structure should support it!)

julian-seward1:
Looks great!  Thanks for the rework.  The hybrid sort/bin-search above threshold is a neat idea, and the transition point (16) sounds about right.

Only minor comments are:

* I'm still concerned about cost of the `num_vregs` calculation, per comment yesterday

* In the final commit message, could you add an ultra-short indication of the rough win?  Eg, "reduces insn count / malloc count by up to / approximately X% on <some test case>" ?

lib/src/data_structures.rs
```diff
@@ -1085,6 +1142,168 @@ impl RegVecsAndBounds {
...
+            slots: Vec::with_capacity(vreg_capacity),
+            overlay: SmallVec::new(),
+        }
+    }
+
+    /// Is the overlay sorted?
+    fn is_overlay_sorted(&self) -> bool {
+        // Use the SmallVec spill-to-heap threshold as a threshold for "large
+        // enough to sort"; this has the effect of amortizing the cost of
+        // sorting alaong with the cost of copying out to heap memory, and also
```
bjorn3, line 1176:
```suggestion
        // sorting along with the cost of copying out to heap memory, and also
```

lib/src/data_structures.rs
```diff
@@ -1085,6 +1142,168 @@ impl RegVecsAndBounds {
...
+    /// Allocate a reg-usage mapper with the given predicted vreg capacity.
+    pub(crate) fn new(vreg_capacity: usize) -> RegUsageMapper {
+        RegUsageMapper {
+            slots: Vec::with_capacity(vreg_capacity),
+            overlay: SmallVec::new(),
+        }
+    }
+
+    /// Is the overlay sorted?
+    fn is_overlay_sorted(&self) -> bool {
```
bnjbvr, line 1173: I thought this would say whether the overlay would be sorted, not if it needed sorting. Should this be renamed to `is_overlay_large_enough_to_sort` (as suggested by below comment) or `overlay_must_be_sorted`?

lib/src/data_structures.rs
```diff
@@ -1085,6 +1142,168 @@ impl RegVecsAndBounds {
...
+
+    /// Finish updates to the overlay, sorting if necessary.
+    pub(crate) fn finish_overlay(&mut self) {
+        if self.is_overlay_sorted() {
+            // Sort stably, so that later updates continue to come after earlier
+            // ones.
+            self.overlay.sort_by_key(|pair| pair.0);
+            // Remove duplicates by collapsing runs of same-vreg pairs down to
+            // the last one.
+            if self.overlay.len() > 0 {
```
bnjbvr, line 1196: This check could go at the top of this function and cause an early return if the overlay is empty, removing one level of indent here.

All indent can even be removed if the `is_overlay_sorted` checked is reversed too.

lib/src/data_structures.rs
```diff
@@ -1085,6 +1142,168 @@ impl RegVecsAndBounds {
...
+                for i in 1..self.overlay.len() {
+                    let this_vreg = self.overlay[i].0;
+                    if this_vreg != last_vreg {
+                        out += 1;
+                    }
+                    if i != out {
+                        self.overlay[out] = self.overlay[i];
+                    }
+                    last_vreg = this_vreg;
+                }
```
bnjbvr, line 1208: Shouldn't the size of the overlay array be changed after collapsing happened? Otherwise it's not removing anything here...

lib/src/data_structures.rs
```diff
@@ -781,41 +848,31 @@ impl fmt::Debug for VirtualReg {
...
-    pub fn apply_defs_or_uses(&mut self, map: &Map<VirtualReg, RealReg>) {
-        if let Some(vreg) = self.as_virtual_reg() {
-            if let Some(rreg) = map.get(&vreg) {
-                debug_assert!(rreg.get_class() == vreg.get_class());
-                *self = rreg.to_reg();
-            } else {
-                panic!("Reg::apply_defs_or_uses: no mapping for {:?}", self);
-            }
-        }
+    /// Apply a vreg-rreg mapping to a Reg.  This used for registers used in
```
bnjbvr, line 851: pre-existing: This `is` used for registers used ...

lib/src/inst_stream.rs
```diff
@@ -100,7 +100,7 @@ fn map_vregs_to_rregs<F: Function>(
     insts_to_add: &Vec<InstToInsertAndPoint>,
     iixs_to_nop_out: &Vec<InstIx>,
     reg_universe: &RealRegUniverse,
-    has_multiple_blocks_per_frag: bool,
+    _has_multiple_blocks_per_frag: bool,
```
bnjbvr, line 103: Could we remove this parameter instead?

lib/src/data_structures.rs
cfallin: Good suggestion, thanks!

lib/src/data_structures.rs
cfallin: Yes, good catch, thanks!

cfallin:
Updated, thanks!

PR #56, opened 2020-04-30
A semantics-level clean-up following implementation of `RangeFrag` co…

julian-seward1:
…mpression

in the backtracking allocator:

* split `RangeFrag` into `RangeFrag` and `RangeFragMetrics`

* in many places where before a vector of `RangeFrag`s was available, now we
  have two vectors of the same length, one of `RangeFrag`s and the other of
  `RangeFragMetrics`.

* add extensive commenting explaining this

* remove from LSRA all complexity to do with `RangeFrag` metrics, which was in
  any case unused

* as a ridealong, change some uses of the `RangeFrag` vector named `fenv` to
  `frag_env` so as to be consistent throughout the codebase, and a bit more
  descriptive.

### DIFF
--- a/lib/src/analysis_data_flow.rs
+++ b/lib/src/analysis_data_flow.rs
@@ -6,9 +6,10 @@ use std::fmt;
 
 use crate::analysis_control_flow::CFGInfo;
 use crate::data_structures::{
-    BlockIx, InstIx, InstPoint, Map, Queue, RangeFrag, RangeFragIx, RangeFragKind, RealRange,
-    RealRangeIx, RealReg, RealRegUniverse, Reg, RegSets, RegUsageCollector, RegVecBounds, RegVecs,
-    RegVecsAndBounds, Set, SortedRangeFragIxs, SpillCost, TypedIxVec, VirtualRange, VirtualRangeIx,
+    BlockIx, InstIx, InstPoint, Map, Queue, RangeFrag, RangeFragIx, RangeFragKind,
+    RangeFragMetrics, RealRange, RealRangeIx, RealReg, RealRegUniverse, Reg, RegSets,
+    RegUsageCollector, RegVecBounds, RegVecs, RegVecsAndBounds, Set, SortedRangeFragIxs, SpillCost,
+    TypedIxVec, VirtualRange, VirtualRangeIx,
 };
 use crate::sparse_set::SparseSet;
 use crate::union_find::{ToFromU32, UnionFind};
@@ -781,6 +782,7 @@ fn get_range_frags_for_block<F: Function>(
     rvb: &RegVecsAndBounds,
     out_map: &mut Map<Reg, Vec<RangeFragIx>>,
     out_frags: &mut TypedIxVec<RangeFragIx, RangeFrag>,
+    out_frag_metrics: &mut TypedIxVec<RangeFragIx, RangeFragMetrics>,
 ) {
     //println!("QQQQ --- block {}", bix.show());
     // BEGIN ProtoRangeFrag
@@ -846,7 +848,7 @@ fn get_range_frags_for_block<F: Function>(
 
     // The generated RangeFrags are initially are dumped in here.  We
     // group them by Reg at the end of this function.
-    let mut tmp_result_vec = SmallVec::<[(Reg, RangeFrag); 32]>::new();
+    let mut tmp_result_vec = SmallVec::<[(Reg, RangeFrag, RangeFragMetrics); 32]>::new();
 
     // First, set up `state` as if all of `livein` had been written just
     // prior to the block.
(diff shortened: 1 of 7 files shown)


lib/src/analysis_data_flow.rs
```diff
@@ -1009,15 +1013,17 @@ fn get_range_frags_for_block<F: Function>(
...
         //println!("QQQQ post: leftover: {}", (r,frag).show());
-        tmp_result_vec.push((*r, frag));
+        tmp_result_vec.push((*r, frag, frag_metrics));
     }
 
     // Copy the entries in `tmp_result_vec` into `out_map` and `outVec`.
     // TODO: do this as we go along, so as to avoid the use of a temporary
     // vector.
-    for (r, frag) in tmp_result_vec {
+    assert!(out_frags.len() == out_frag_metrics.len());
```
bnjbvr, line 1025: nit: `debug_assert_eq` here?

lib/src/analysis_data_flow.rs
```diff
@@ -1094,7 +1105,8 @@ pub fn get_range_frags<F: Function>(
     }
 
     info!("    get_range_frags: end");
-    (result_map, result_frags)
+    assert!(result_frags.len() == result_frag_metrics.len());
```
bnjbvr, line 1108: ditto

lib/src/analysis_data_flow.rs
```diff
@@ -1105,11 +1117,13 @@ pub fn get_range_frags<F: Function>(
 fn merge_range_frags_slow(
     frag_ix_vec_per_reg: &Map<Reg, Vec<RangeFragIx>>,
     frag_env: &TypedIxVec<RangeFragIx, RangeFrag>,
+    frag_metrics_env: &TypedIxVec<RangeFragIx, RangeFragMetrics>,
     cfg_info: &CFGInfo,
 ) -> (
     TypedIxVec<RealRangeIx, RealRange>,
     TypedIxVec<VirtualRangeIx, VirtualRange>,
 ) {
+    assert!(frag_env.len() == frag_metrics_env.len());
```
bnjbvr, line 1126: ditto (etc for other appearances in this PR)

lib/src/analysis_data_flow.rs
```diff
@@ -1409,8 +1424,8 @@ pub fn merge_range_frags(
                 continue 'per_frag_loop;
             }
             // This frag isn't Local (standalone) so we have process it the slow way.
```
bnjbvr, line 1426: nit: we have `to` process it

lib/src/bt_commitment_map.rs
```diff
@@ -87,10 +87,10 @@ impl fmt::Debug for CommitmentMap {
     }
 }
 impl CommitmentMap {
-    pub fn show_with_fenv(&self, _fenv: &TypedIxVec<RangeFragIx, RangeFrag>) -> String {
+    pub fn show_with_frag_env(&self, _frag_env: &TypedIxVec<RangeFragIx, RangeFrag>) -> String {
         //let mut frags = TypedIxVec::<RangeFragIx, RangeFrag>::new();
         //for fix in &self.frag_ixs {
-        //  frags.push(fenv[*fix]);
+        //  frags.push(frag_env[*fix]);
         //}
```
bnjbvr, line 94: Should we remove code that's commented out? It's deader than dead code, in the sense that we don't even know if it compiles when we'll re-enable it.

lib/src/bt_main.rs
```diff
@@ -801,9 +824,10 @@ pub fn alloc_main<F: Function>(
     let rlr_env: TypedIxVec<RealRangeIx, RealRange> = analysis_info.1;
     let mut vlr_env: TypedIxVec<VirtualRangeIx, VirtualRange> = analysis_info.2;
     let mut frag_env: TypedIxVec<RangeFragIx, RangeFrag> = analysis_info.3;
-    let _liveouts: TypedIxVec<BlockIx, SparseSet<Reg>> = analysis_info.4;
-    let est_freqs: TypedIxVec<BlockIx, u32> = analysis_info.5;
-    let inst_to_block_map: InstIxToBlockIxMap = analysis_info.6;
+    let frag_metrics_env: TypedIxVec<RangeFragIx, RangeFragMetrics> = analysis_info.4;
+    let _liveouts: TypedIxVec<BlockIx, SparseSet<Reg>> = analysis_info.5;
```
bnjbvr, line 828: is it time that `analysis_info` gets named fields? (hint: yes) :)

lib/src/data_structures.rs
```diff
@@ -1824,27 +1769,37 @@ impl RangeFrag {
...
-        last: InstPoint,
-        count: u16,
-    ) -> Self {
-        debug_assert!(f.block_insns(bix).len() >= 1);
-        debug_assert!(f.block_insns(bix).contains(first.iix));
-        debug_assert!(f.block_insns(bix).contains(last.iix));
-        RangeFrag::new_multi_block(f, bix, first, last, count)
-    }
+    // LSRA has a commented-out use of this.  We can reinstate it if
+    // the caller gets reinstated.
```
bnjbvr, line 1779: I now recall that I was using this function to avoid running into the assertions that now live in `new_with_metrics`; so from this point, LSRA can use `new` instead, and not have the assertions. Can you please remove this ctor? (but keep LSRA's code; I need to think if i keep it or not)

lib/src/analysis_data_flow.rs
julian-seward1: There are a lot of plain `assert`s in the analysis, data structures and bt code.  So I'll stick with `assert` for now.  I do try not to have them in hot loops, and I keep watch for assert-y costs when looking at profiles.  At some point we might think to have an assertion audit and/or policy discussion.

PR #59, opened 2020-05-06
Allow `block_succs` to return a Cow[BlockIx] instead of a Vec[BlockIx]

bnjbvr:
In addition with a wasmtime PR (about to be posted), this helps reducing the number of memory allocations during compilations.

Also contains a small utility improvement: ability to not print the function's output onto stdout.

### DIFF
--- a/bin/test_framework.rs
+++ b/bin/test_framework.rs
@@ -5,7 +5,7 @@ use arbitrary::Arbitrary;
 use regalloc::*;
 use std::collections::HashSet;
 
-use std::fmt;
+use std::{borrow::Cow, fmt};
 
 use crate::validator::{validate, Context as ValidatorContext, RegRef};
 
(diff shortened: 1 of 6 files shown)


bin/test_framework.rs
```diff
@@ -2178,9 +2178,9 @@ impl regalloc::Function for Func {
     }
 
     /// Get CFG successors: indexed by block, provide a list of successor blocks.
-    fn block_succs(&self, block: BlockIx) -> Vec<BlockIx> {
+    fn block_succs(&self, block: BlockIx) -> Cow<[BlockIx]> {
         let last_insn = self.blocks[block].start.plus(self.blocks[block].len - 1);
-        self.insns[last_insn].get_targets()
+        Cow::Owned(self.insns[last_insn].get_targets())
```
sunfishcode, line 2183: Could this use `Box<[BlockIx]>` instead of `Cow`? You can convert a `Vec` to a `Box` via [`into_boxed_slice()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_boxed_slice).

bjorn3: At the Cranelift side this returns `Cow::Borrowed`.

bnjbvr: Right, we're trying to avoid the memory allocation that comes with the cloning, and giving ownership isn't a viable option either (fwiw, this memory allocation was in the top 3 of the DHAT profile).

PR #60, opened 2020-05-11
Mark fn get_sanitized_reg_uses_for_func as inline-always.  Reduces in…

julian-seward1:
…sn count

for one spill-heavy input by around 1.4%.  Surprisingly high, but I think that
inlining means that the result triple never needs to be created.
No functional change.

### DIFF
--- a/lib/src/analysis_data_flow.rs
+++ b/lib/src/analysis_data_flow.rs
@@ -435,7 +435,7 @@ pub fn get_sanitized_reg_uses_for_func<F: Function>(
 // convenience interface
 
 // ==== EXPORTED ====
-#[inline(never)]
+#[inline(always)]
 pub fn does_inst_use_def_or_mod_reg(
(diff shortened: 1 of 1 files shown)


bjorn3:
Does `#[inline]` have the same effect?

julian-seward1:
I think it is more limited and doesn't work across crate or module boundaries (one or the other).

julian-seward1:
I re-pushed because the commit message mentions the wrong function.

bjorn3:
When optimizations are enabled `#[inline]` and `#[inline(always)]` are considered equal during partitioning: there will be a copy in each codegen unit that uses them. When optimizations are disabled however `#[inline]` will be considered almost equal to a non inline function for partitioning purposes.

https://github.com/rust-lang/rust/blob/fb5615a4771ea3d54256f969dc84d2dfd38d812c/src/librustc_middle/mir/mono.rs#L107

PR #63, opened 2020-05-19
Data flow analysis: use more idiomatic Rust code

bnjbvr:
This avoids a few map lookups by using map methods that returns the
existing entries, sometimes reusing them in place (instead of inserting
a copy in the same slot).

This is a very slight reduction in number of executed instructions when compiling regex-rs.wasm (0.2%) or big.clif (<0.01%).


lib/src/analysis_data_flow.rs
```diff
@@ -1345,13 +1347,13 @@ pub fn merge_range_frags(
     let mut result_real = TypedIxVec::<RealRangeIx, RealRange>::new();
     let mut result_virtual = TypedIxVec::<VirtualRangeIx, VirtualRange>::new();
 
-    /* BEGIN per_reg_loop */
-    'per_reg_loop: for (reg, all_frag_ixs_for_reg) in frag_ix_vec_per_reg.iter() {
```
julian-seward1, line 1392: Please reinstate these loop names.  I find it useful for reading comprehension that the continue statements carry the name of the loop.

lib/src/analysis_data_flow.rs
```diff
@@ -1486,65 +1491,67 @@ pub fn merge_range_frags(
...
                             ix_left = triples_len;
                         }
                         if ix_left < triples_len {
                             assert!(ix_left == 0 || triples[ix_left - 1].2 < *b);
                         }
+
                         // ix2 plays the same role as in the quadratic version.  ix_left and
                         // ix_right are not used after this point.
                         let mut ix2 = ix_left;
-                        'loop_over_entries_for_b: loop {
```
julian-seward1, line 1556: Ditto here (and everywhere)

lib/src/data_structures.rs
```diff
@@ -1887,19 +1887,15 @@ impl fmt::Debug for SortedRangeFragIxs {
 }
 
 impl SortedRangeFragIxs {
-    pub fn check(&self, fenv: &TypedIxVec<RangeFragIx, RangeFrag>) {
-        let mut ok = true;
+    pub(crate) fn check(&self, fenv: &TypedIxVec<RangeFragIx, RangeFrag>) {
+        #[cfg(debug_assertions)]
```
julian-seward1, line 1891: I believe this changes the failure behaviour.  `panic!` always terminates the process, regardless of build configuration.  So the old version would always have paniced given invalid input.  The new version only panics when given invalid input when it is built with `#[cfg(debug_assertions)]`.  If you remove `#[cfg(debug_assertions)]` then I think the behaviour will be unchanged.

lib/src/analysis_data_flow.rs
bnjbvr: Well, for one I find these take over the attention, making me think that there might continue/break statements from an inner loop to an outer loop. So I was looking for this, and didn't find such statements. As a reader, this is quite confusing, because you expect something unusual to happen with respect to continue statements, but it does not in fact. And since the continue statements do the regular thing, that is, continue to the innermost loop, I found the code to be more readable without the loop names: the loop names represent unnecessary complexity/pollution of the code.

With this in mind, would you be ok removing the loop names? I'm happy to keep comments where the loop start and end, if that helps.

lib/src/data_structures.rs
bnjbvr: Alright, I explicitly wanted to make this debug-only; I'll keep it in release mode too as well since this seems to matter.

lib/src/analysis_data_flow.rs
julian-seward1: Yes, fair point actually .. it hadn't occurred to me that the presence of such names indicates non-local continues/breaks.  So yes ok, remove them then.

PR #64, opened 2020-05-19
/!\ Transform RegUsageMapper into a trait

bnjbvr:
This allows several implementations of RegUsageMapper to co-exist
without the client being aware of it.

This also renames the existing RegUsageMapper to VrangeRegUsageMapper,
since it tracks the mappings for the whole lifetime of given virtual
ranges. It is moved to its own file "reg_maps.rs", to reduce the size of the bag of All The Things also known as data_structures.rs. LSRA's own implementation of RegUsageMapper could also go there.

A few functions need to get a type parameter, and they get monomorphized
at compile-time.

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -25,6 +25,7 @@ mod checker;
 mod data_structures;
 mod inst_stream;
 mod linear_scan;
+mod reg_maps;
 mod sparse_set;
 mod union_find;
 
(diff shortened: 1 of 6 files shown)


lib/src/reg_maps.rs
```diff
@@ -0,0 +1,287 @@
...
+use smallvec::SmallVec;
+use std::mem;
+
+/// This data structure holds the mappings needed to map an instruction's uses, mods and defs from
+/// virtual to real registers.
+///
+/// It remembers the sets of mappings (of a virtual register to a real register) over time, based
+/// on vranges and their allocations.
+#[derive(Debug)]
+pub struct VrangeRegUsageMapper {
```
cfallin, line 11: Perhaps add a comment here (or a block comment at the top of the file) indicating what's special about a "vrange" RUM vs. any other RUM? I.e., when would I use this one in particular? And maybe at least a TODO to foreshadow what other RUM implementations may be built in the future.

lib/src/lib.rs
```diff
@@ -148,9 +149,21 @@ pub use crate::data_structures::RegClassInfo;
 
 pub use crate::data_structures::RegUsageCollector;
 
-// A structure for providing mapping results for a given instruction.
+// A trait for providing mapping results for a given instruction.
```
cfallin, line 152: Make this a doc-comment? Also, pre-existing, but we should probably add some more details here -- something about how this provides information about virtual -> real mappings for any sort of register reference (use, def or mod) in an instruction, to be used when editing instructions after performing allocation; and point to the place (`map_regs()`) where this is used in the public interface.

lib/src/reg_maps.rs
bnjbvr: Thanks, this got me thinking about the name of the data structure as well, which I myself find less than ideal. So this implementation is to be used when the virtual ranges are precisely tracked, as the backtracking allocator does; LSRA will use imprecise ranges, so it'll use a different implementation based on precise mentions (i.e. use/mod/def) at instruction points. I'm tempted to rename this "PreciseRegUsageMapper" instead, or "PreciseVrangeRegUsageMapper". Since this is internal anyways, I'll keep it as is right now, and will propose a follow-up if we agree on a better name.

PR #66, opened 2020-05-20
BT: clean up and document some interactions between the coalescing an…

julian-seward1:
…alysis

and the spill slot allocator.

The coalescing analysis computes, amongst other things, equivalence classes of
VirtualRanges.  Two VirtualRanges will be assigned the same equivalence class
if there is a move instruction that transfers a value from one range to the
other.  The equivalence classes created are the transitive closure of this
pairwise relation.

The main allocation loop tries to assign the same register to all the
VirtualRanges in an equivalence class.  Similarly, the spill slot allocator
tries to allocate the same spill slot to all the VirtualRanges in an
equivalence class.  In most cases they are successful, and so the moves
between those VirtualRanges will later disappear.  However, the complete story
is not quite so simple.

It is only safe to assign the VirtualRanges in the same equivalence class to a
single register or spill slot if those VirtualRanges are non-overlapping.
That is, if their overall collection of RangeFrags is disjoint.  If two such
VirtualRanges overlapped, then they could be carrying different values, and so
they would need separate registers or spill slots.

Most of the time, these equivalence classes are indeed internally
non-overlapping.  But that's just luck -- that's how the input VCode mostly
is.  The coalescing analysis *doesn't* check for overlaps within an
equivalence class, so it can be the case that the members of an equivalence
class overlap.  The users of that information -- the main allocation loop and
the spill slot allocator -- currently check for, and handle, such situations.
So the generated allocation is correct.

It does, however, cause imprecision and unnecessary spilling, and, in the main
allocation loop, slightly increased evictions.  Worse, though, is that these
invariants are all undocumented.

This patch makes the following changes:

* It documents them; the above comment (more or less) is added at the start of bt_coalescing_analysis.rs.

* In bt_coalescing_analysis.rs, there is a half-hearted effort made to avoid
  creating equivalence classes whose elements (VirtualRanges) overlap.  This
  is done by doing an overlap check on two VirtualRanges connected by a move,
  and not merging their equivalence classes if they overlap.  That helps, but
  it doesn't completely avoid the problem because there might be overlaps
  between other members (VirtualRanges) of the about-to-be-merged equivalence
  classes.

  Fixing this properly would entail doing a complete overlap check for the two
  equivalence classes before merging.  That sounds expensive, though.  Hence
  this partial fix.

* In the spill slot allocator (bt_spillslot_allocator.rs), the search space
  (meaning, the available spill slots) checked in each call to
  `alloc_spill_slots` has been greatly reduced.  Now, only the
  highest-numbered 8 slots are searched for spaces.  That has two effects:

  - it avoids lots of pointless repeated checking of low-numbered spill slots,
    that long ago became full(ish) and are unlikely to be able to take any new
    VirtualRanges

  - more subtly, it interacts with the question of whether or not each
    VirtualRange equivalence class is internally overlapping.  When no
    overlaps are present, the spill slot allocator guarantees to find a slot
    which is free for the entire equivalence class, which is the ideal
    solution.  When there are overlaps present, the allocator is forced to
    allocate at least some of the VirtualRanges in the class to different
    slots.  By restricting the number of slots it can choose to 8 (+ extras if
    it needs them), we reduce the tendency for the VirtualRanges to be
    assigned a large number of different slots, which in turn reduces the
    amount of spilling in the end.

Of course the "proper" fix for all this is to fix the coalescing analysis so
as only to build non-internally-overlapping VirtualRange equivalence classes.

* Minor change: fix some misleading naming (vlrXX vs vlrixXX) in
  `do_coalescing_analysis`.

* Minor change: a bunch of commented out debug printing has been removed.

Some test cases take longer to compile, some less, but in no case is the insn
count difference greater than 0.1%.

For `joey_big.clif`, which spills heavily, static instruction counts are:

  BEFORE: 8151 spills, 8829 reloads, 337 total spill slots,
          69280 moves coalesced out at the spill level

  AFTER:  6477 spills, 7157 reloads, 356 total spill slots,
          70948 moves coalesced out at the spill level

so a significant reduction in static spill/reload counts, and only a modest
increase in the number of spill slots used.

### DIFF
--- a/lib/src/bt_coalescing_analysis.rs
+++ b/lib/src/bt_coalescing_analysis.rs
@@ -1,7 +1,30 @@
+//! Analysis in support of copy coalescing for the backtracking allocator.
+//!
+//! This detects and collects information about all copy coalescing
+//! opportunities in the incoming function.  It does not use that information
+//! at all -- that is for the main allocation loop and the spill slot allocator
+//! to do.
+//!
+//! Coalescing analysis creates 4 pieces of information:
+//!
+//! * a map from `VirtualRangeIx` to a set of `Hint`s (see below) which state a
+//!   preference for which register that range would prefer to be allocated to.
+//!
+//! * equivalence class groupings for the virtual ranges.  Two virtual ranges
+//!   will be assigned the same equivalence class if there is a move instruction
+//!   that transfers a value from one range to the other.  The equivalence
+//!   classes created are the transitive closure of this pairwise relation.
+//!
+//! * a simple mapping from instruction index to bool, indicating those
+//! instructions that are moves between virtual registers, and that have been
+//! used to construct the equivalence classes above.
+//!
+//! * a mapping from virtual registers to virtual ranges.  This is really
+//!   produced as a side-effect of computing the above three elements, but is
+//!   useful in its own right and so is also returned.
+
 #![allow(non_snake_case)]
 #![allow(non_camel_case_types)]
-
-//! Copy-coalescing analysis for the backtracking allocator.
 
 use log::{debug, info, log_enabled, Level};
 use smallvec::{smallvec, SmallVec};
@@ -15,12 +38,48 @@ use crate::union_find::{ToFromU32, UnionFind, UnionFindEquivClasses};
 use crate::Function;
 
 //=============================================================================
-// Analysis in support of copy coalescing.
 //
-// This detects and collects information about all copy coalescing
-// opportunities in the incoming function.  It does not use that information
-// at all -- that is for the main allocation loop to do.
+// *** Some important comments about the interaction between this coalescing
+// *** analysis, the main allocation loop and the spill slot allocator.
+//
+// The main allocation loop tries to assign the same register to all the
+// VirtualRanges in an equivalence class.  Similarly, the spill slot allocator
+// tries to allocate the same spill slot to all the VirtualRanges in an
+// equivalence class.  In most cases they are successful, and so the moves
+// between those VirtualRanges will later disappear.  However, the complete
+// story is not quite so simple.
+//
+// It is only safe to assign the VirtualRanges in the same equivalence class
+// to a single register or spill slot if those VirtualRanges are
+// non-overlapping.  That is, if their overall collection of RangeFrags is
+// disjoint.  If two such VirtualRanges overlapped, then they could be
+// carrying different values, and so they would need separate registers or
+// spill slots.
+//
+// Most of the time, these equivalence classes are indeed internally
+// non-overlapping.  But that's just luck -- that's how the input VCode mostly
+// is.  The coalescing analysis *doesn't* properly check for overlaps within an
+// equivalence class, so it can be the case that the members of an equivalence
+// class overlap.  The users of that information -- the main allocation loop
+// and the spill slot allocator -- currently check for, and handle, such
+// situations.  So the generated allocation is correct.
+//
+// It does, however, cause imprecision and unnecessary spilling, and, in the
+// main allocation loop, slightly increased evictions.
+//
+// The "proper" fix for all this would be to fix the coalescing analysis so as
+// only to build non-internally-overlapping VirtualRange equivalence classes.
+// However, that sounds expensive.  Instead there is a half-hearted effort
+// made to avoid creating equivalence classes whose elements (VirtualRanges)
+// overlap.  This is done by doing an overlap check on two VirtualRanges
+// connected by a move, and not merging their equivalence classes if they
+// overlap.  That helps, but it doesn't completely avoid the problem because
+// there might be overlaps between other members (VirtualRanges) of the
+// about-to-be-merged equivalence classes.
 
+//=============================================================================
+// Coalescing analysis: Hints
+//
 // A coalescing hint for a virtual live range.  The u32 is an arbitrary
 // "weight" value which indicates a relative strength-of-preference for the
 // hint.  It exists because a VLR can have arbitrarily many copy
@@ -147,7 +206,8 @@ pub fn do_coalescing_analysis<F: Function>(
     //  println!("QQQQ rreg r{:?} -> rlrixs {:?}", rreg, rlrixs);
     //}
 
-    // Range end checks for VRegs
+    // Range end checks for VRegs.  The XX means either "Last use" or "First
+    // def", depending on the boolean parameter.
     let doesVRegHaveXXat
     // `xxIsLastUse` is true means "XX is last use"
     // `xxIsLastUse` is false means "XX is first def"
@@ -178,7 +238,7 @@ pub fn do_coalescing_analysis<F: Function>(
       None
     };
 
-    // Range end checks for RRegs
+    // Range end checks for RRegs.  XX has same meaning as above.
     let doesRRegHaveXXat
     // `xxIsLastUse` is true means "XX is last use"
     // `xxIsLastUse` is false means "XX is first def"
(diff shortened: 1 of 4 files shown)


lib/src/data_structures.rs
```diff
@@ -1964,6 +1967,44 @@ impl SortedRangeFrags {
             frags: SmallVec::<[RangeFrag; 4]>::new(),
         }
     }
+
+    pub fn overlaps(&self, other: &Self) -> bool {
```
cfallin, line 1971: A general comment: it might be a good idea to establish a convention from now on that the small but algorithmically non-trivial helper functions have reasonable unit-test suites. The algorithm below (in `ovleraps()`) is pretty evidently correct, and we have "integration-test" coverage of a sort with the whole-RA tests, including fuzzing and the Cranelift use-case, but I I think it would increase confidence somewhat if we have some more targeted tests too.

(I know the tree/UF/set/... data structures are pretty well-tested, which is great! I'm thinking more about the mid-level building blocks, and even the sections of some of the long analysis functions; e.g. computation of each intermediate map or list could be split into a separate helper and tested with a bunch of weird cases.)

lib/src/bt_spillslot_allocator.rs
```diff
@@ -215,18 +240,29 @@ impl SpillSlotAllocator {
...
-        // individually.
+        // Pass 1: find a slot which can take all VirtualRanges in `vlrix`s
+        // eclass when tested individually.
         //
-        // 1a: search existing slots
+        // Pass 1a: search existing slots
+        let search_start_slotno: u32 = {
+            // We will only search from `search_start_slotno` upwards.  See
+            // block comment above for significance of the value `8`.
+            let window = 8;
```
cfallin, line 250: Do we want to make this a tunable? We have some infra now for `Options`; perhaps this and other heuristics could be plumbed up to the top level so that we can play with them later if we want to tune, or come up with aggressive/not-aggressive packages of defaults, etc.

lib/src/data_structures.rs
julian-seward1: Yes .. testing the mid level helpers and building blocks would be a good thing to bear in mind.

I'd also like to add larger tests to the regalloc.rs standalone test suite (the `.rat` files), in particular tests that stress coalescing and spilling more, and are more representative of vcode translated from SSA sources.  Currently the test suite is very poor in this respect.  Problem is we have no way to compile C directly the `.rat` pseudo-architecture, and translating inputs by hand isn't feasible -- I translated a 3-way quicksort function and that took a whole day.

lib/src/bt_spillslot_allocator.rs
julian-seward1: I don't think there's much point.  Most small and medium functions won't even use 8 spill slots, even on arm64, except maybe across calls.  In my tests with `joey_big.clif`, reducing this number caused the number of required spill slots to rise only slowly: 8 requires 356 slots, infinity requires 337.  And increasing the number causes more spilling, which I believe is due to the "fragmentation of `VirtualRange` groups across multiple spill slots" phenomenon documented in the patch.

lib/src/bt_coalescing_analysis.rs
```diff
@@ -19,8 +19,68 @@ use crate::Function;
...
+//
+// * a map from `VirtualRangeIx` to a set of `Hint`s (see below) which state a
+//   preference for which register that range would prefer to be allocated to.
+//
+// * equivalence class groupings for the virtual ranges.  Two virtual ranges
+//   will be assigned the same equivalence class if there is a move instruction
+//   that transfers a value from one range to the other.  The equivalence
+//   classes created are the transitive closure of this pairwise relation.
+//
+// * a simple mapping from insn index to bool, indicating those instructions
```
bnjbvr, line 35: nit: please fully expand "instruction" in comment

lib/src/bt_coalescing_analysis.rs
```diff
@@ -280,25 +340,30 @@ pub fn do_coalescing_analysis<F: Function>(
...
-                    hints[vlrDst].push(Hint::SameAs(vlrSrc, block_eef));
-                    vlrEquivClassesUF.union(vlrDst, vlrSrc);
-                    is_vv_boundary_move[iix] = true;
-                    // Reduce the total cost, and hence the spill cost, of
-                    // both `vlrSrc` and `vlrDst`.  This is so as to reduce to
-                    // zero, the cost of a VLR whose only instructions are its
-                    // v-v boundary copies.
-                    debug!("QQQQ reduce cost of {:?} and {:?}", vlrSrc, vlrDst);
-                    decVLRcosts.push((vlrSrc, vlrDst, 1 * block_eef));
+                let mb_vlrixSrc = doesVRegHaveXXat(/*xxIsLastUse=*/ true, rSrcV, iix);
```
bnjbvr, line 343: a very peculiar function name... Can you clarify it, btw, please?

lib/src/bt_coalescing_analysis.rs
```diff
@@ -280,25 +340,30 @@ pub fn do_coalescing_analysis<F: Function>(
...
+                let mb_vlrixDst = doesVRegHaveXXat(/*xxIsLastUse=*/ false, rDstV, iix);
+                if mb_vlrixSrc.is_some() && mb_vlrixDst.is_some() {
+                    let vlrixSrc = mb_vlrixSrc.unwrap();
+                    let vlrixDst = mb_vlrixDst.unwrap();
+                    // Per block comment at top of file, make a half-hearted
+                    // attempt to avoid creating equivalence classes with
+                    // internal overlaps.  Note this can't be completely
+                    // effect as presently implemented.
```
bnjbvr, line 351: Not sure, but... `this can't be completely effect as presently implemented` doesn't sound right to me. Did you mean `effective`?

lib/src/bt_main.rs
```diff
@@ -669,12 +669,21 @@ pub fn alloc_main<F: Function>(
...
+            // start.
+            //
+            // However, it may be the case that there is some other VLR which
+            // is in the same equivalence class as `curr_vlr`, but is not a
+            // direct neighbour, and which has already been assigned a
+            // register.  We really ought to take those into account too, as
+            // the least-preferred candidates.  Hence we need to iterate over
+            // the equivalence class and "round up the secondary candidates."
+            //
+            // Note that the equivalance class might contain VirtualRanges
```
bnjbvr, line 681: nit: equivalence

lib/src/bt_coalescing_analysis.rs
```diff
@@ -19,8 +19,68 @@ use crate::Function;
 //
 // This detects and collects information about all copy coalescing
 // opportunities in the incoming function.  It does not use that information
-// at all -- that is for the main allocation loop to do.
+// at all -- that is for the main allocation loop and the spill slot allocator
+// to do.
+//
+// Coalescing analysis creates 4 pieces of information:
```
bnjbvr, line 25: General comment for this block: can you make this a doc comment for the whole file? with `//!`?

lib/src/bt_coalescing_analysis.rs
```diff
@@ -280,25 +340,30 @@ pub fn do_coalescing_analysis<F: Function>(
...
-                    vlrEquivClassesUF.union(vlrDst, vlrSrc);
-                    is_vv_boundary_move[iix] = true;
-                    // Reduce the total cost, and hence the spill cost, of
-                    // both `vlrSrc` and `vlrDst`.  This is so as to reduce to
-                    // zero, the cost of a VLR whose only instructions are its
-                    // v-v boundary copies.
-                    debug!("QQQQ reduce cost of {:?} and {:?}", vlrSrc, vlrDst);
-                    decVLRcosts.push((vlrSrc, vlrDst, 1 * block_eef));
+                let mb_vlrixSrc = doesVRegHaveXXat(/*xxIsLastUse=*/ true, rSrcV, iix);
+                let mb_vlrixDst = doesVRegHaveXXat(/*xxIsLastUse=*/ false, rDstV, iix);
```
bnjbvr, line 344: I don't know how costful is this function, but if it is, would it make sense to lazily compute the value of `mb_vlrixDst` only if the src one `is_some()`? Then, there's probably a way to do something like `if let Some(vlrixSrc) = doesVRegHaveXXat(...) { if let Some(vlrixDst) = doesVregHaveXXat(...) { ... } }`, and avoid the weird "check `is_some()` then `unwrap`" pattern (which might be imply an extra check that the option is not empty).

lib/src/data_structures.rs
```diff
@@ -1964,6 +1967,44 @@ impl SortedRangeFrags {
...
+                    continue;
+                }
+                _ => {
+                    // There's no unique "next frag" -- either they are
+                    // identical, or they overlap.  So we're done.
+                    return true; // "there's an overlap"
+                }
+            }
+        }
+        /*NOTREACHED*/
```
bnjbvr, line 2006: nit: this is not needed, because I'm pretty sure the Rust compiler will warn you if you insert code here, saying it is unreachable.

lib/src/data_structures.rs
```diff
@@ -1964,6 +1967,44 @@ impl SortedRangeFrags {
...
+                // We made it to the end of one (or both) vectors without
+                // finding any conflicts.
+                return false; // "no overlaps"
+            }
+            let f1 = &frags1[c1];
+            let f2 = &frags2[c2];
+            match cmp_range_frags(f1, f2) {
+                Some(Ordering::Less) => {
+                    c1 += 1;
+                    continue;
```
bnjbvr, line 1993: nit: here and below, the continue statements aren't needed, since this is where the control flow naturally goes

lib/src/data_structures.rs
```diff
@@ -1964,6 +1967,44 @@ impl SortedRangeFrags {
...
+        let mut c2 = 0;
+        loop {
+            if c1 >= n1 || c2 >= n2 {
+                // We made it to the end of one (or both) vectors without
+                // finding any conflicts.
+                return false; // "no overlaps"
+            }
+            let f1 = &frags1[c1];
+            let f2 = &frags2[c2];
+            match cmp_range_frags(f1, f2) {
```
bnjbvr, line 1990: Fun fact: i did implement exactly this code for LSRA at some point, to detect the first InstPoint that would overlap two different virtual ranges. One small speedup i got, was to use a binary search to find the start index (`c1` and `c2`), when one of the virtual ranges had many fragments.

lib/src/bt_coalescing_analysis.rs
julian-seward1: Hmm, interesting point.  I might come back to this in some other PR -- this one is really about reducing spilling, not about speedups.

For cases like `if x.is_some() { .. x.unwrap() .. }` I tend to assume that LLVM's inlining and CSEing will remove the duplicate (implied) check.  Similarly for the case you mentioned, I tend to assume that LLVM will move (by doing PRE) the computation of `mb_vlrixDst` to happen after the `mb_vlrixSrc.is_some()` is done.  But maybe I expect too much of it.

lib/src/data_structures.rs
julian-seward1: Interesting .. again, somewhere to come back to.  Also because the "sparse array, attempt 2" implementation I am thinking of will also involve parallel traversals through sorted arrays of Things.

PR #68, opened 2020-05-28
BT: avoid construction of "no-evict" sets

julian-seward1:
When BT searches for a usable register to hold a VLR that belongs to an
equivalence class of VLRs (as determined by the coalescing analysis), it is
important not to choose for eviction, any other member of that class.  That
would undermine the whole point of coalescing.  It wouldn't lead to incorrect
allocations, though.

Until now, BT constructed, for each VLR, a SparseSet holding all the VLRs in
the same equivalence class, and passed it to `find_evict_set`.
`find_evict_set` then checks this set and will not request eviction of any
member in it.  The sets are constructed by consulting the
`UnionFindEquivClasses` created by coalescing analysis.

This works, but is wasteful: if there is a large equivalence class (of VLRs)
containing N elements, then the same no-evict set, of size N, will be
constructed N times.  This is quadratically bad in the worst case.

This patch removes all of the set construction and testing.  Instead,
`UnionFindEquivClasses` is enhanced so that it can directly answer the
question "are these two items in the same equivalence class?"  Then, a
suitable closure that calls the new function is passed to `find_evict_set`,
and that uses it to make the same evict/no-evict decision that it previously
did via set-membership tests of the no-evict set.

A closely related ridealong fix, is that all of the methods of `struct
SpillCost` are now marked as inline-always.  They were called a lot from
`find_evict_set` and so the inlining is a significant win.

Overall, for huge functions (`joey_big.clif`) this reduces insn count by 6.4%
and the number of malloc calls by 25.4% (!).

### DIFF
--- a/lib/src/bt_main.rs
+++ b/lib/src/bt_main.rs
@@ -120,7 +120,7 @@ impl PerRealReg {
 // adding their spill costs to `running_cost`.  Return false if, for one of
 // the 4 reasons documented below, the traversal has been abandoned, and true
 // if the search completed successfully.
-fn search_commitment_tree(
+fn search_commitment_tree<IsAllowedToEvict>(
     // The running state, threaded through the tree traversal.  These
     // accumulate ranges and costs as we traverse the tree.  These are mutable
     // because our caller (`find_evict_set`) will want to try and allocate
@@ -134,9 +134,12 @@ fn search_commitment_tree(
     // The RangeFrag we want to accommodate.
     pair_frag: &RangeFrag,
     spill_cost_budget: &SpillCost,
-    do_not_evict: &SparseSetU<[VirtualRangeIx; 16]>,
+    allowed_to_evict: &IsAllowedToEvict,
     vlr_env: &TypedIxVec<VirtualRangeIx, VirtualRange>,
-) -> bool {
+) -> bool
+where
+    IsAllowedToEvict: Fn(VirtualRangeIx) -> bool,
+{
     let mut stack = SmallVec::<[u32; 32]>::new();
     assert!(tree.root != AVL_NULL);
     stack.push(tree.root);
@@ -173,8 +176,8 @@ fn search_commitment_tree(
             if !vlr_to_evict.spill_cost.is_less_than(spill_cost_budget) {
                 return false;
             }
-            // Maybe our caller asked us not to evict this one.
-            if do_not_evict.contains(vlrix_to_evict) {
+            // Maybe our caller doesn't want us to evict this one.
+            if !allowed_to_evict(vlrix_to_evict) {
                 return false;
             }
             // Ok!  We can evict the current node.  Update the running state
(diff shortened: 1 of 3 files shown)


lib/src/bt_main.rs
```diff
@@ -120,7 +120,7 @@ impl PerRealReg {
 // adding their spill costs to `running_cost`.  Return false if, for one of
 // the 4 reasons documented below, the traversal has been abandoned, and true
 // if the search completed successfully.
-fn search_commitment_tree(
+fn search_commitment_tree<ALLOWED_TO_EVICT>(
```
bnjbvr, line 123: nit: it's unusual to use this casing for a type parameter. In addition to casing, i would suggest a slightly different name: what about `IsAllowedToEvict` or `AllowedToEvictFn` (Fn for function)?

lib/src/bt_main.rs
```diff
@@ -252,12 +255,15 @@ impl PerRealReg {
     //   non-infinite-cost eviction candidates.  This is by design (so as to
     //   guarantee that we can always allocate spill/reload bridges).
     #[inline(never)]
-    fn find_evict_set(
+    fn find_evict_set<ALLOWED_TO_EVICT>(
```
bnjbvr, line 258: ditto here

lib/src/bt_main.rs
```diff
@@ -828,9 +829,15 @@ pub fn alloc_main<F: Function>(
...
 
-            let mb_evict_info: Option<(SparseSetU<[VirtualRangeIx; 4]>, SpillCost)> = per_real_reg
-                [rregNo]
-                .find_evict_set(curr_vlrix, &empty_Set_VirtualRangeIx, &vlr_env);
+            let mb_evict_info: Option<(SparseSetU<[VirtualRangeIx; 4]>, SpillCost)> =
+                per_real_reg[rregNo].find_evict_set(
+                    curr_vlrix,
+                    // We pass a closure that ignores its arg and returns `true`.
+                    // Meaning, "we are not specifying any particular
+                    // can't-be-evicted VLRs in this call.
```
bnjbvr, line 837: nit: missing closing `"` here, or remove the opening one

lib/src/union_find.rs
```diff
@@ -345,6 +348,32 @@ impl<T: ToFromU32> UnionFind<T> {
...
+    // class.  If either falls outside the "universe", returns `None`.
+    pub fn in_same_equivalence_class(&self, item1: T, item2: T) -> Option<bool> {
+        let mut item1num = ToFromU32::to_u32(item1) as usize;
+        let mut item2num = ToFromU32::to_u32(item2) as usize;
+        // If either item is outside our "universe", say we don't know.
+        if item1num >= self.heads.len() || item2num >= self.heads.len() {
+            return None;
+        }
+        // Ensure that `item1num` and `item2num` both point at class leaders.
+        if (self.heads[item1num] & 0x8000_0000) == 0 {
```
bnjbvr, line 362: I am just discovering about this code, but my first impression here is that this raw constant should be named and used in the entire file. Even better, there could be an always-inlined function `is_class_leader(item) -> bool` that would encapsulate this constant, if not `get_class_leader(item) -> item`. Something to keep in mind for later.

PR #69, opened 2020-05-29
Rewrite `UnionFind::find` to improve performance.  This is a small re…

julian-seward1:
…cursive

function that is called a lot.  Attempts to replace the recursion with an
explicit stack generated mediocre results, so instead the function has been
unrolled and split into hot and cold variants.  There are further explanatory
comments in-line.  For `joey_big.clif` this reduces the compilation insn count
by almost 1%.

### DIFF
--- a/lib/src/union_find.rs
+++ b/lib/src/union_find.rs
@@ -97,21 +97,100 @@ impl<T: ToFromU32> UnionFind<T> {
     // Find, with path compression.  Returns the index of tree root for the
     // given element.  This is not for external use.  There's no boundary
     // checking since Rust will do that anyway.
-    /*priv*/
-    fn find(&mut self, elem: u32) -> u32 {
-        let elem_parent_or_size: i32 = self.parent_or_size[elem as usize];
-        if elem_parent_or_size < 0 {
+    //
+    // This was initially implemented using a recursive function.  However,
+    // this function gets called a lot, and the recursion led to significant
+    // expense.  Attempts to replace the recursion with an explicit stack
+    // didn't give much speedup.  Hence the following scheme, which retains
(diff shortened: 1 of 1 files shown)


lib/src/union_find.rs
```diff
@@ -97,21 +97,92 @@ impl<T: ToFromU32> UnionFind<T> {
...
+    // those 4 iterations don't complete the traversal back to the root, it
+    // calls onwards to `find_slow`, which recurses.  The idea is that `find`
+    // handles the majority of the cases and can always be inlined, and we
+    // hand off the remaining cases to `find_slow` which will never be inlined
+    // (and hence will not interfere with the inlining of this function).
+    #[inline(always)]
+    fn find(&mut self, elem0: u32) -> u32 {
+        // Handle up to 4 steps up the tree in-line.
+        let elem0_parent_or_size: i32 = self.parent_or_size[elem0 as usize];
+        if elem0_parent_or_size < 0 {
```
bnjbvr, line 119: Can you introduce an always-inlined helper function `parent_or_root(elem: u32) -> Option<u32>` (returns the parent index if there's one, `None` if it's a root), that's used here instead of the raw check? This would make this code easier to understand. Or at least a helper function `is_root(u32) -> bool` that encapsulates the fact that `< 0` means root?

julian-seward1: In the end I didn't do this, because I didn't want to give rustc the opportunity to mess up performance by failing to inline the helper for whatever reason.  What I did do though is to add a comment immediately before `fn find` about the meaning of negative values.  It's also documented further up the file but a reminder at `fn find` is no bad thing.

PR #71, opened 2020-06-01
A few memory enhancements

bnjbvr:
This series of commits reduce the number of memory allocations during compilation with Cranelift:

- first commit uses the default implementation of Clone for Set, which is more efficient (as it avoids reallocations) and has the same effect.
- second commit reduces the number of allocations in get_range_frags, by: flushing RangeFrag as they're created instead of accumulating them in a Vec flushed at the end; reusing memory across block iterations. Some of this work might be undone by Julian's future patches, but it's a direct improvement here.
- The last commit (breaking change) requires that add_uses/add_defs are given Slices of Reg/WritableReg instead of Sets, which reduces the number of memory allocations, when combined with a symmetrical patch in Cranelift.

Each commit has been tested for performance separately and did improve all the metrics (icount, number of malloc'd bytes and blocks), when tested on joey-big/joey-medium/regex-rs.wasm test cases. For the latter, it's a 9% icount decrease, 19% block allocations decrease, 10% total of allocated bytes decrease.

### DIFF
--- a/lib/src/data_structures.rs
+++ b/lib/src/data_structures.rs
@@ -956,6 +956,7 @@ impl fmt::Debug for SpillSlot {
 pub struct RegUsageCollector<'a> {
     pub reg_vecs: &'a mut RegVecs,
 }
+
 impl<'a> RegUsageCollector<'a> {
     pub fn new(reg_vecs: &'a mut RegVecs) -> Self {
         Self { reg_vecs }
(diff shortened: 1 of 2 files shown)


lib/src/data_structures.rs
```diff
@@ -33,6 +33,7 @@ pub type Map<K, V> = FxHashMap<K, V>;
 // Sets of things
 
 // Same comment as above for FxHashMap.
+#[derive(Clone)]
```
julian-seward1, line 36: Umm, yes.  Now I wonder why I didn't do this to begin with :-)

PR #72, opened 2020-06-04
Remove all `Map` uses from `get_range_frags` and `get_range_frags_for…

julian-seward1:
…_block`

This patch replaces them with vectors which are indexed by "unified reg index",
which is explained in comments in the patch.  There are two such vectors:

* `out_map`: this contains a SmallVec of RangeFragIxs for each Reg.  This is
  the overall result of `get_range_frags`.

* `state`: this is used in each call to `get_range_frags_for_block`, to keep
  track of the ProtoRangeFrag state for each register as we move through the
  block.  It replaces the previous Map based mapping.

One complication is that we no longer have available, the keys (Regs) for
`state`.  Those are needed so as to know which entries to flush at the end of
`get_range_frags_for_block`.  To solve this, the `state` indices that each
call to `get_range_frags_for_block` visits, are stored in a new vector
`visited`, and that is used to drive the end-of-function flushing.

To avoid repeatedly allocating `visited`, it is allocated/owned by
`get_range_frags`, and passed to each call to `get_range_frags_for_block`.

`get_range_frags` also now has to determine the number of virtual regs, and
the RegClass for each, in order that `get_range_frags_for_block` and later
`merge_range_frags` can map from "unified register index" back to the original
Reg values.

`merge_range_frags` is adjusted to take input in the new format (vectors
instead of maps).

enum RegClass acquires a new value, ::INVALID, to facilitate incremental
construction of the abovementioned mapping from virtual regs to their classes.

Ridealong and minor mods:

* new function `Reg::get_index_u32`, so as to avoid usize-to-u32-and-back
  casts when working with `Reg::get_index` -- these functions are now hot.

* `SparseSetU::upgrade`: use a stragically-placed `reserve` call to avoid
  a few allocations later.

* `SparseSetU::contains`: use an existing function to loop over the small
  array, rather than duplicating code.  Also, make this inline-always; it
  is hot.

Overall, this reduces the number of blocks allocated by 15% (joey_small.clif)
to 19% (joey_big.clif) and the insn count by 0% (joey_big.clif) to 3.5%
(joey_med.clif).

### DIFF
--- a/lib/src/analysis_data_flow.rs
+++ b/lib/src/analysis_data_flow.rs
@@ -1,14 +1,14 @@
 //! Performs dataflow and liveness analysis, including live range construction.
 
 use log::{debug, info, log_enabled, Level};
-use smallvec::SmallVec;
+use smallvec::{smallvec, SmallVec};
 use std::cmp::min;
 use std::fmt;
 
 use crate::analysis_control_flow::CFGInfo;
 use crate::data_structures::{
-    BlockIx, InstIx, InstPoint, Map, Point, Queue, RangeFrag, RangeFragIx, RangeFragKind,
-    RangeFragMetrics, RealRange, RealRangeIx, RealReg, RealRegUniverse, Reg, RegSets,
+    BlockIx, InstIx, InstPoint, Point, Queue, RangeFrag, RangeFragIx, RangeFragKind,
+    RangeFragMetrics, RealRange, RealRangeIx, RealReg, RealRegUniverse, Reg, RegClass, RegSets,
     RegUsageCollector, RegVecBounds, RegVecs, RegVecsAndBounds, SortedRangeFragIxs,
     SortedRangeFrags, SpillCost, TypedIxVec, VirtualRange, VirtualRangeIx,
 };
@@ -776,6 +776,7 @@ pub fn calc_livein_and_liveout<F: Function>(
 /// The running state of this function is a map from Reg to ProtoRangeFrag.  Only Regs that
 /// actually appear in the Block (or are live-in to it) are mapped.  This has the advantage of
 /// economy, since most Regs will not appear in (or be live-in to) most Blocks.
+#[derive(Clone)]
 struct ProtoRangeFrag {
     /// The InstPoint in this Block at which the associated Reg most recently became live (when
     /// moving forwards though the Block).  If this value is the first InstPoint for the Block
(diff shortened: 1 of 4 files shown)


lib/src/analysis_data_flow.rs
```diff
@@ -828,59 +934,40 @@ fn get_range_frags_for_block<F: Function>(
...
-
-        // And use the new RangeFragIx.
-        out_map.entry(r).or_insert_with(|| Vec::new()).push(new_fix);
-    };
+    // Invariants for the preallocated storage:
+    //
+    // * `visited` is always irrelevant (and cleared) at the start
+    //
+    // * `state` always has size (# real regs + # virtual regs).  However, all
+    //   its entries should be `None` in between calls here.
```
bnjbvr, line 942: Can you add a `#[cfg(debug_assertions)]`-only block that checks that all the entries are set to None?

lib/src/analysis_data_flow.rs
```diff
@@ -983,83 +1079,198 @@ fn get_range_frags_for_block<F: Function>(
...
+            implied_num_vregs = r_ix + 1;
+            vreg_classes.resize(implied_num_vregs, RegClass::INVALID);
+        }
+        // rustc 1.43.0 appears to have problems avoiding duplicate bounds checks for
+        // `vreg_classes[r_ix]`; hence give it a helping hand here.
+        let vreg_classes_ptr = &mut vreg_classes[r_ix];
+        if *vreg_classes_ptr == RegClass::INVALID {
+            *vreg_classes_ptr = r.get_class();
+        } else {
+            assert!(*vreg_classes_ptr == r.get_class());
```
bnjbvr, line 1194: nit: `assert_eq!` will give better error messages, in case of errors.

lib/src/analysis_data_flow.rs
```diff
@@ -983,83 +1079,198 @@ fn get_range_frags_for_block<F: Function>(
...
     assert!(rvb.is_sanitized());
-    let mut result_map = Map::<Reg, Vec<RangeFragIx>>::default();
+
+    // In order that we can work with unified-reg-indices (see comments
+    // above), we need to know (1) how many virtual regs there are and (2) the
+    // `RegClass` for each.  That info is collected in a single pass here.  In
+    // principle regalloc.rs's user could tell us (1), but as yet the
+    // interface does not make that possible.
+    let mut vreg_classes = Vec::</*vreg index,*/ RegClass>::new();
+    let mut implied_num_vregs = 0;
```
bnjbvr, line 1172: Maybe this could just be called "num_vregs"?

lib/src/analysis_data_flow.rs
```diff
@@ -983,83 +1079,198 @@ fn get_range_frags_for_block<F: Function>(
...
-        debug!("{:<3?}   {:?}", RangeFragIx::new(n), frag);
-        n += 1;
+    assert!(state.len() == num_regs);
+    assert!(result_map.len() == num_regs);
+    assert!(vreg_classes.len() == num_virtual_regs);
+    // This is pretty cheap (once per fn) and any failure will be catastrophic
+    // since it means we may have forgotten some live range fragments.
+    // Hence `assert!` and not `debug_assert!`.
+    for state_elem in &state {
+        assert!(state_elem.is_none());
```
bnjbvr, line 1246: Ah, you actually did it ;) Can you move this assertion at the end of `get_range_frags_for_block` instead?

lib/src/sparse_set.rs
```diff
@@ -70,6 +70,7 @@ where
             SparseSetU::Small { card, arr } => {
                 assert!(*card == A::size());
                 let mut set = FxHashSet::<A::Item>::default();
+                set.reserve(A::size());
```
bnjbvr, line 73: Is there a `with_capacity` method that could be used to create the hashset with the expected capacity, in the previous line?

lib/src/analysis_data_flow.rs
```diff
@@ -983,83 +1079,198 @@ fn get_range_frags_for_block<F: Function>(
...
+            // This can't happen.  `r` is in `liveout`, but this implies
+            // that it is neither defined in the block nor present in
+            // `livein`.
+            None => panic!("get_range_frags_for_block: fail #3"),
+            Some(ref pf) => {
+                // `r` is written (or modified), either literally or by virtue
+                // of being present in `livein`, and may or may not
+                // subsequently be read -- we don't care, because it must be
+                // read "after" the block.  Create a `LiveOut` or `Thru` frag
+                // accordingly.
```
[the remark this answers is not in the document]
julian-seward1: Fixed, for all comments in `get_range_frags{,_for_block}` and anything else the patch touches.

lib/src/analysis_data_flow.rs
julian-seward1: Well, I considered that, but didn't do it because I was concerned it would make debug builds (even more) unusably slow.  It would cost (# regs * # basic blocks).  Instead, the array is checked just once, in `get_range_frags`, after we have processed all blocks.  This is a weaker test than what you propose, for sure, but I am hoping it is good enough :-)

lib/src/analysis_data_flow.rs
julian-seward1: See comment above; in short, moving it to the end of `get_range_frags_for_block` would change the cost from once-per-function to once-per-block, which would be a huge perf hit.  So I'd prefer not to.

lib/src/sparse_set.rs
julian-seward1: Yeah, I wondered about that too.  Rust-lang.org's documentation for `HashSet` mentions that `with_capacity` exists, but when I tried that the compiler complained it doesn't exist.  Maybe that's because I am actually using `FxHashSet` here?  But I thought `FxHashSet` was identical to `HashSet` except it has a different hasher?  So I'm generally confused; I don't know why the compiler complained.

PR #73, opened 2020-06-04
A tiny bit more smallvec-ification/preallocation.  When applied on to…

julian-seward1:
…p of

PR72, this reduces the number of malloc calls by 6% (joey_small), 7%
(joey_med) and 48% (joey_big).  It is neutral or a small win for all of those
in both instruction count terms and total bytes allocated.


lib/src/avl_tree.rs
```diff
@@ -61,7 +61,10 @@ impl<T: Clone> AVLTree<T> {
     // Create a new tree and its associated storage pool.  This requires knowing
     // the default item value.
     pub fn new(default: T) -> Self {
-        let pool = vec![];
+        // We have relatively few trees -- one for each rreg and one per spill
+        // slot -- but they tend to get big.  So pre-allocate a few entries so
+        // as to save a few reallocs later.
```
bnjbvr, line 66: This comment reflects insider knowledge of the backtracking allocator, but this is only part of the story; LSRA uses AVL trees for a different purpose.

Two things:

- LSRA uses the AVL tree to keep a list of sorted intervals to be allocated later; can you quickly check it doesn't regress performance of LSRA either?
- can you remove the part of the comment that's specific to backtracking?

PR #74, opened 2020-06-08
IR snapshots + benchmarking with Criterion

bnjbvr:
This PR implements several things:

- a snapshotting mechanism that allows any regalloc.rs embedder to make a snapshot of their IR, and serialize it, if the `enable-serde` Cargo feature is enabled.
- reloading a snapshot from the minira CLI util, so it can be replayed independently of the compiler.
- as a proof of concept, I made snapshots of the 3 "joey" CLIF functions that are compiled with Cranelift at the best speed level.
- and I added [criterion](https://github.com/bheisler/criterion.rs) benchmarking on top of this, to be able to get some interesting automated reports (both in CLI and HTML reports) of how the performance evolves over time. (See commit message for more details.)


benches/compare.rs
```diff
@@ -0,0 +1,87 @@
...
+        algorithm: Algorithm::LinearScan(LinearScanOptions::default()),
+    };
+
+    let bt_opts = Options {
+        run_checker: false,
+        algorithm: Algorithm::Backtracking(BacktrackingOptions::default()),
+    };
+
+    {
+        let joey_smallest = read("./snapshots/joey_smallest.bin");
```
cfallin, line 27: Are there better names we can use for these tests (more descriptive)? Also, do we know anything about the license situation? At the very least we probably need a copyright statement if we're embedding binary code (more or less). Alternately, maybe we leave the test suite out of this PR and keep a separate repo or copy somewhere. Alternately alternately, maybe we take a few tests (e.g. in the cfallin/cranelift-new-backend-tests repo) with source and `Makefile`s, and write a little shell script to compile them and snapshot the IR, given an external wasmtime checkout?

lib/src/snapshot.rs
```diff
@@ -0,0 +1,310 @@
...
+        }
+        for reg_def in insn.reg_defs.iter_mut() {
+            if let Some(vreg) = reg_def.to_reg().as_virtual_reg() {
+                *reg_def =
+                    Writable::from_reg(maps.get_def(vreg).expect("missing alloc for def").to_reg());
+            }
+        }
+    }
+
+    fn gen_spill(
```
cfallin, line 247: For snapshotting purposes, do we need to actually generate "real" instructions (i.e., that capture the parameter values and record the spill/reload op) here? Since we will not interpret the code, and the register allocator core has no way to query whether an instruction is a spill/reload later, perhaps we can just return a generic `IRInst::Nop` or similar for `gen_spill`/`gen_reload`.

It doesn't really affect our benchmarking too much (removes a few `Vec` allocations, I suppose), but it simplifies the interface implementation here...

benches/compare.rs
bnjbvr: I've asked about licensing.

> Alternately, maybe we leave the test suite out of this PR and keep a separate repo or copy somewhere.

This would invalidate all the benchmarking work, since the benchmarks require to have the binary tests checked in. Minimizing friction is important for better adoption, and requiring to clone a different repository that contains the benchmarks seems a bit unfortunate. Ideally, we'd have a small number of very important benchmarks checked in, and a less restricted set of benchmarks available in another repository.

> Alternately alternately, maybe we take a few tests (e.g. in the cfallin/cranelift-new-backend-tests repo) with source and Makefiles, and write a little shell script to compile them and snapshot the IR, given an external wasmtime checkout?

We can, but this really becomes outside the scope of this PR, which is a first step in the good direction, in my opinion.

lib/src/snapshot.rs
bnjbvr: I didn't come as far as this, but: it would be possible to retrieve the number of spills/reloads in the current form, which is not the best proxy but is one proxy for the resulting generated code's quality. So maybe it's just a matter of making all those public.

benches/compare.rs
cfallin: Yep, agreed, the best option is to keep the test-cases in the repo, and I don't want to force a bunch of unnecessary work here either!

@jgouly, would it be possible to ask you to check in the `.wasm` test files you've given us in a subdirectory, with whatever copyright notice is needed, so we have a clear chain of origin for the binary IR snapshots here?

lib/src/snapshot.rs
cfallin: Ah, yeah, that's true; IMHO it's fine to leave any potential optimizations here for later.

bjorn3:
Should we really permanently increase the repo size by 3.3 MB not just once, but every time the serialization format changes? Maybe the benchmarks could be moved to a separate crate that depends on Cranelift and then check in the wasm files once?

PR #76, opened 2020-06-10
Interface (stubbed out) for safepoint/reference-type support.

cfallin:
This PR is just a sketch of how I propose that safepoint support might
look from the API perspective; the safeopints are not actually respected
yet. Possible starting point for real safepoint/reference-type/GC
support.

Feel free to merge if you'd prefer, or just start from this patch if that's easier!


cfallin:
Closing this PR (old/outdated now that we have real reftypes under review).

PR #79, opened 2020-06-27
Bugfix and API adjustment for reference types support.

cfallin:
This PR fixes a small bug in the reference-type support from #80, and makes a minor API adjustment.

In particular, this bugfix ensures that the "reffyness" computation is as precise as we need it to be (by propagating reffyness only to specific ranges associated with registers, rather than all of them, since e.g. a given rreg might be reffy at one program-point but not another).

Tested by integrating into Cranelift (PR over there will be created once this lands) and then into SpiderMonkey from there (patch also pending).

I'll plan to add checker support to verify stackmaps so that we can fuzz this and have a bit more confidence in it.

### DIFF
--- a/lib/src/analysis_data_flow.rs
+++ b/lib/src/analysis_data_flow.rs
@@ -8,10 +8,10 @@ use std::fmt;
 use crate::analysis_control_flow::CFGInfo;
 use crate::data_structures::{
     BlockIx, InstIx, InstPoint, MoveInfo, MoveInfoElem, Point, Queue, RangeFrag, RangeFragIx,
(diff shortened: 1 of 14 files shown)


lib/src/analysis_data_flow.rs
```diff
@@ -1870,13 +1870,40 @@ pub fn compute_reg_to_ranges_maps<F: Function>(
     }
 }
 
-// Collect info about registers that are connected by moves:
+// Collect info about registers (and optionally Virtual/RealRanges) (and optionally
```
julian-seward1, line 1873: nit: "(and optionally Virtual/RealRanges)" appears twice.

cfallin: D'oh, it seems I was holding `vim` wrong; fixed!

PR #80, opened 2020-07-01
Initial implementation of reftypes support in the allocator (currentl…

julian-seward1:
…y, BT only).  A summary of

the changes is:

* If the client wants stackmap/reftypes support, it has to provide three extra pieces of
  information:
  - the register class holding reftyped values, on the target
  - the set of VirtualRegs that hold reftyped values (that it created)
  - the set of instructions for which stackmaps (and, implicitly, safepoints) are required.
  In return it is given:
  - for each safepoint instruction, the set of stack slots holding reftyped values
  - a mapping from old safepoint instruction indicies to the final insn indices, since the
    allocator will both add and remove instructions

* For RealRanges and VirtualRanges, there is an extra bool to indicate whether the range is
  reftyped.

* Some analysis work needed by both the reftype dataflow analysis (see next point) and BT's
  coalescing machinery, has been refactored so as to avoid doing it twice.  Specifically,
  computation of maps from RealRegs to RealRanges and from VirtualRegs to VirtualRanges.

* A dataflow analysis phase used only when reftype support is requested.  From the initial set
  of reftyped virtual registers (not ranges), compute the set of all VirtualRanges and
  RealRanges that are reftyped.  This is done in the new file analysis_reftypes.rs.  Note this
  code is buggy and Chris has a fix which will land later.  It is also inefficient and will need
  to be rewritten once it is more stable.

* In the spill slot allocator, keep track of which slots at which ranges are allocated to a
  reftyped VirtualRange.  Add a mechanism to query each slot's tree at some instruction so as to
  see whether that slot contains a reftyped value at that instruction.

* For each instruction at which the client requests stackmaps, do the following:

  - Add instructions that spill live reftyped RealRegs onto the stack, just before the insn, and
    add spill slots as necessary to hold them.

  - Similarly, add instructions to reload, after the instruction, the subset of spilled values
    which remain live after the call.

  The effect is to force all reftyped values onto the stack at each safepoint, which allows them
  to be updated by the GC that implicitly runs "in the middle of the safepoint", and then to
  reload those values into RealRegs as required.

  - Finally, then, consult the StackSlotAllocator's state at each safepoint instruction.  This
    gives the set of slots that are reftyped, which is the basis for making stackmaps.

* In bt_coalescing_analysis.rs, fn do_coalescing_analysis, there has been a drive-by cleanup:
  the two strangely named closures `doesVRegHaveXXat` and `doesRRegHaveXXat`, which tool a bool
  indicating the meaning of the `XX`, have each been split in half, and the bool removed.  This
  gives four closures, no runtime interpretation of booleans, and simpler to follow code.

* In the case where reftype support is not requested, the changes are performance neutral.  In
  fact there is some extra overhead, but that is more than cancelled out by a small
  representational improvement in CommitmentMap.

### DIFF
--- a/lib/src/analysis_reftypes.rs
+++ b/lib/src/analysis_reftypes.rs
@@ -1,0 +1,142 @@
+//! Performs a simple taint analysis, to find all live ranges that are reftyped.
+
+use crate::data_structures::{
+    MoveInfo, MoveInfoElem, RangeId, RealRange, RealRangeIx, RegClass, RegToRangesMaps, TypedIxVec,
+    VirtualRange, VirtualRangeIx, VirtualReg,
+};
+use crate::sparse_set::SparseSet;
+
+pub fn do_reftypes_analysis(
+    // From dataflow/liveness analysis.  Modified by setting their is_ref bit.
+    rlr_env: &mut TypedIxVec<RealRangeIx, RealRange>,
+    vlr_env: &mut TypedIxVec<VirtualRangeIx, VirtualRange>,
+    // From dataflow analysis
+    reg_to_ranges_maps: &RegToRangesMaps,
+    move_info: &MoveInfo,
+    // As supplied by the client
+    reftype_class: RegClass,
+    reftyped_vregs: &Vec<VirtualReg>,
+) {
+    // The game here is: starting with `reftyped_vregs`, find *all* the VirtualRanges and
+    // RealRanges to which that refness can flow, via instructions which the client's `is_move`
+    // function considers to be moves.
+
+    // We have `move_info`, which tells us which regs (both real and virtual) are connected by
+    // moves.  However, that's not directly useful -- we need to know which *ranges* are
+    // connected by moves.  So first, convert `move_info` into a set of range-pairs.
+
+    let mut range_pairs = Vec::<(RangeId, RangeId)>::new(); // (DST, SRC)
+
+    for MoveInfoElem {
+        dst: dst_reg,
+        src: src_reg,
+        iix: _,
+        est_freq: _,
+    } in &move_info.moves
+    {
+        debug_assert!(dst_reg.get_class() == src_reg.get_class());
+
+        // Don't waste time processing moves which can't possibly be of reftyped values.
+        if dst_reg.get_class() != reftype_class {
+            continue;
+        }
+
+        // This is kinda tiresome because of the differing representations, but .. construct the
+        // cartesian product of the range indicies for both the source and destination of the
+        // move.
+
+        let dst_reg_ix = dst_reg.get_index();
+        let src_reg_ix = src_reg.get_index();
+        match (dst_reg.is_real(), src_reg.is_real()) {
+            (true, true) => {
+                // R <- R
+                for dst_ix in &reg_to_ranges_maps.rreg_to_rlrs_map[dst_reg_ix] {
+                    for src_ix in &reg_to_ranges_maps.rreg_to_rlrs_map[src_reg_ix] {
+                        range_pairs.push((RangeId::new_real(*dst_ix), RangeId::new_real(*src_ix)));
+                    }
+                }
+            }
+            (true, false) => {
+                // R <- V
+                for dst_ix in &reg_to_ranges_maps.rreg_to_rlrs_map[dst_reg_ix] {
+                    for src_ix in &reg_to_ranges_maps.vreg_to_vlrs_map[src_reg_ix] {
+                        range_pairs
+                            .push((RangeId::new_real(*dst_ix), RangeId::new_virtual(*src_ix)));
+                    }
+                }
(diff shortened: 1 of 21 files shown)


lib/src/bt_main.rs
```diff
@@ -368,6 +382,162 @@ fn print_RA_state(
...
+    // SpiderMonkey use case and where `iix` is a call insn.  That's because SM's calling
+    // convention has no callee saved registers.  Hence "real regs written by `iix`" will be
+    // "all real regs" and so Safter will be empty.  And Sbefore is in any case pretty small.
+    //
+    // (/me thinks ..) hmm, if Safter is empty, then what is the point of dumping Sbefore on the
+    // stack before the GC?  For r in Sbefore, either r is the only reference to some object, in
+    // which case there's no point in presenting that ref to the GC since r is dead after call,
+    // or r isn't the only ref to the object, in which case some other ref to it must exist
+    // elsewhere in the stack, and that will keep the object alive.  Maybe this needs a rethink.
+    // Maybe the spills before the call should be only for the set Safter?
```
cfallin, line 439: Yes, the conclusion here makes sense to me: we only really need to spill what's in `Safter`. I think this is equivalent to the following change (or falls out of it, anyway): rather than "Sbefore is all real regs that are live at `iix.u` that are reftypes", we have "Sbefore is all real regs that are live strictly beyond `iix.u`". We can thus get `Safter` by checking each range that we examine and testing whether `iix.d` is also included in the range. That may or may not be more efficient than subtracting out the real regs written by the instruction (?).

In any case, I'm happy to get this landed now and reconsider efficiencies later, because as you say, it's a moot point for SM anyway...

lib/src/inst_stream.rs
```diff
@@ -76,14 +76,112 @@ impl InstToInsert {
...
+// safepoint spill/reload instructions can be added after "normal" spill/reload instructions
+// have been created, and it doesn't interact with the logic to create those "normal"
+// spill/reload instructions.
+//
+// In the worst case scenario, a value could be reloaded at R, immediately spilled at SB, then
+// possibly modified in memory at the safepoint proper, reloaded at RA, and spilled at S.  That
+// is considered to be an unlikely scenario, though.
+
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ExtPoint {
```
cfallin, line 127: A high-level design question: what is the reason for separating `Point` and `ExtPoint`, and should we document this (or merge if appropriate)?

I can imagine at least a few reasons: e.g., most of the allocator should not place anything at the `SB` / `RA` points, and using only `Point` ensures this at the type level. Also perhaps there are performance reasons (packed representation of `InstAndExtPoint`), though I imagine we could simply steal another bit from the instruction index if needed.

But there's also a small complexity cost imposed by having two separate notions of "instruction point", understanding of which depends on understanding of the phase structure, and if we later move reftype handling earlier in the RA pipeline, or spill-before for other (as yet unimagined) reasons, we'd have to plumb `ExtPoints` in everywhere. If we really depend on the fact that a `SB` / `RA` point can't come from the pre-reftypes algorithm, perhaps we could assert that somewhere instead?

Anyway, I'm fine either way -- just want to make sure we have a good reason for this split!

lib/src/analysis_reftypes.rs
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    // Now, finally, compute the fixpoint resulting from repeatedly mapping `reftyped_ranges`
+    // through `range_pairs`. XXXX this looks dangerously expensive .. reimplement.
+    //
+    // Later .. this is overkill.  All that is needed is a DFS of the directed graph in which
+    // the nodes are the union of the RealRange(Ixs) and the VirtualRange(Ixs), and whose edges
+    // are exactly what we computed into `range_pairs`.  This graph then needs to be searched
+    // from each root in `reftyped_ranges`.
```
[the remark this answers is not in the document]
julian-seward1: Files as https://github.com/bytecodealliance/regalloc.rs/issues/81.

lib/src/bt_main.rs
julian-seward1: We can come back to this.  (And should.)  Not easy to reason about.

lib/src/inst_stream.rs
julian-seward1: `ExtPoint` is used only for insn editing/merging (a peripheral role, in the grand scheme of things).  I guess it seemed conceptually cleaner to have dataflow and liveness analysis remain using `Point`, in the sense that I didn't want to have `SB` and `RA` available there when they have no meaning, and possibly having to have dynamic checks to that effect.

That all said, until we actually get to generating spills/reloads, even `Point::R` and `S` have no role and should not appear.  So I don't really have a consistent story here.

PR #84, opened 2020-07-13
Add checker and fuzzing support for reftypes.

### LINKED ISSUE #47: Make the checker aware of backtracking improvements
The checker isn't aware of a few backtracking improvements: compressed ranges, as well as spill slot coalescing.

A few tests are disabled because of this; we should make sure that we'd reenable those (in bin/minira.rs) once the checker has been fixed.

cfallin:
This PR adds:
* Checker support for spillslot coalescing (closes #47).
* Fuzzing-infrastructure support for generating programs with reference type defs and uses and safepoints.
* Checker support for validating stackmaps.

Reftype support in `bt` has been fuzzed up to 10.9B test cases without finding an issue (except for an initial issue in the reftypes checker that I fixed).

## Supporting coalesced spillslots

The key idea of this change is to use "ghost instructions" to
communicate information from the register allocator to the checker.
These instructions have no side-effect, by definition, but communicate a
semantic change to the program state that is relevant to verification:
for example, a move from one virtual register to another, when both
virtual registers live in the same spillslot (a "V-V move").

These ghost instructions exist only as edit-list entries; they are never
actually reified into real instructions and inserted, nor does the
machine backend have to have any special support to generate them. They
are just conceptual.

It turns out that to support spill-slot coalescing (i.e. to successfully
check allocation results with this feature enabled), all we need is to
correctly handle these V-V moves. Other forms of coalescing (e.g., a
move between two virtual registers allocated to the same real register)
are already visible to the checker, because it sees these redundant
moves before they are removed.

In a little more detail, the design-space considered included:

1. Extending the checker to reason about storing multiple virtual
   registers in each storage slot / allocation target (real register or
   spill slot). The problem with this approach is that it is much less
   efficient -- analysis values for each slot at each program point are
   no longer fixed 64-bit things but potentially unbounded sets -- and
   that it also loses information, because we may have two different
   programs that store a different vreg in a slot when arriving at a
   program point via different paths. Fundamentally, we should check the
   move at the point that the sets are union'd (the "ghost instruction")
   and then reason only about the (single) "current" vreg in the slot.

2. Reason about collections of equality constraints by assigning
   symbolic names (tags or whatever) to each def (i.e., SSA values) and
   then gathering move-constraints together. This started to get really
   complex, and seems to require basically a full SSA recovery algorithm
   plus a good part of the allocator's analysis, so it was discarded
   (the checker should be simple and straightforward to maximize the
   confidence we have in it).

3. Reason about reaching definitions, and validate the program by
   showing that the same defs reach each use at each instruction in the
   pre-allocated and post-allocated versions. The problem with this is
   similar to option 1 above: it discards information about paths, and
   so is imprecise.

Fundamentally, it seemed that we could either try to design clever
heuristics to *recover* the information about the elided V-V moves, or
we could just observe them directly before they are elided; so the above
ghost-instruction approach was chosen.

## Checking reftype stackmaps

Next, we add stackmap checking to the checker, and fix some stackmap fuzzing
issues.

The checker for stackmaps works as follows:

- Every analysis value in the checker gets a "reffy" bit. These are
  filled in according to the list of reffy vregs provided by the
  regalloc.

- When stackmap information is given for a safepoint, the checker
  verifies that every slot described as reffy by the stackmap is in fact
  reffy.

- Any reffy spillslot that passes through a safepoint *without* being
  named in the stackmap is set to "undefined". Then, if it is later used
  as a reference (or at all), we will detect that it was in fact live
  but was not named in the stackmap.

Note that these rules do *not* do the obvious thing and construct an
expected stackmap from the spillslot state at the safepoint's program
point, then compare it to the given stackmap. This is because a stackmap
is allowed to elide a reffy spillslot that's dead: we could just as
easily have zeroed it out after its last use, as it semantically no
longer exists in the program; we were just too lazy to clear it. We
don't want to reconstruct liveness information and verify that every
reffy-but-not-described-as-such slot is dead in the checker, so instead
we just mark every reffy slot not named as reffy as undefined. This is
somewhat intuitive when seen a certain way: if the GC (or whatever
stackmap consumer) is unaware of a reference value at a safepoint, it
might be outdated (e.g. if the GC moves every object), so might as well
be undefined.

The stackmap checking appears to truly work because it found the above
dead-spillslot issue quite quickly. After fixing that bug, it ran up to
the following state (10.9B test cases) before I killed it:

`#10943745210: cov: 4362 ft: 19953 corp: 2941 exec/s 1577 oom/timeout/crash: 0/0/0 time: 303772s job: 24310 dft_time: 0`

### DIFF
--- a/bin/fuzzing.rs
+++ b/bin/fuzzing.rs
@@ -67,7 +67,7 @@ impl FuzzingEnv {
 
     /// Returns true whenever a reftyped vreg may be defined.
     fn can_def_reftyped_reg(&self) -> bool {
-        self.ref_regs.len() != (self.num_ref_regs as usize)
+        self.can_use_reftyped_reg() || self.ref_regs.len() != (self.num_ref_regs as usize)
     }
 
     fn def_reg(&mut self, rc: RegClass, u: &mut Unstructured) -> Result<Reg> {
@@ -98,13 +98,29 @@ impl FuzzingEnv {
 
     fn def_reftyped_reg(&mut self, u: &mut Unstructured) -> Result<Reg> {
         debug_assert!(self.can_def_reftyped_reg());
-        let mut index = u16::arbitrary(u)? % self.num_ref_regs;
-        while self.ref_regs.contains(&index) {
-            index = (index + 1) % self.num_ref_regs;
+        if self.ref_regs.len() == 0
+            || (self.ref_regs.len() < (self.num_ref_regs as usize) && bool::arbitrary(u)?)
+        {
+            let mut index = u16::arbitrary(u)? % self.num_ref_regs;
+            while self.ref_regs.contains(&index) {
+                index = (index + 1) % self.num_ref_regs;
+            }
+            self.ref_regs.insert(index);
+            let index = index + self.num_virtual_regs;
+            Ok(Reg::new_virtual(RegClass::I32, index as u32))
+        } else {
+            assert!(self.ref_regs.len() > 0);
+            let list_index = usize::arbitrary(u)? % self.ref_regs.len();
+            let reg_index = self
+                .ref_regs
+                .iter()
+                .skip(list_index)
+                .cloned()
+                .next()
+                .unwrap();
+            let reg_index = reg_index + self.num_virtual_regs;
+            Ok(Reg::new_virtual(RegClass::I32, reg_index as u32))
         }
-        self.ref_regs.insert(index);
-        let index = index + self.num_virtual_regs;
-        Ok(Reg::new_virtual(RegClass::I32, index as u32))
     }
 
     fn mod_reg(&mut self, rc: RegClass, u: &mut Unstructured) -> Result<Reg> {
@@ -165,6 +181,7 @@ impl FuzzingEnv {
             ImmF,
             Copy,
             CopyF,
+            CopyRef,
             BinOp,
             BinOpM,
             BinOpF,
@@ -206,6 +223,9 @@ impl FuzzingEnv {
         if self.can_use_reftyped_reg() && self.can_def_reg(I32) {
             allowed_insts.push(AllowedInst::UseRef);
         }
+        if self.can_def_reftyped_reg() && self.can_use_reftyped_reg() {
+            allowed_insts.push(AllowedInst::CopyRef);
+        }
 
         debug_assert!(!allowed_insts.is_empty());
 
@@ -237,6 +257,11 @@ impl FuzzingEnv {
                     dst: self.def_reg(F32, u)?,
                     src,
                 }
+            }
+            AllowedInst::CopyRef => {
+                let dst = self.def_reftyped_reg(u)?;
+                let src = self.get_reftyped_reg(u)?;
+                Copy { dst, src }
             }
             AllowedInst::BinOp => {
                 let src_left = self.get_reg(I32, u)?;
--- a/bin/fuzzing.rs
+++ b/bin/fuzzing.rs
@@ -20,8 +20,11 @@ const NUM_BLOCK_INSTS: u8 = 8;
 struct FuzzingEnv {
     num_blocks: u8,
     num_virtual_regs: u16,
+    num_ref_regs: u16, // numbered in vreg space above ordinary vregs.
     /// Map of virtual register index to register class. None means the register hasn't been ever defined.
     vregs: HashMap<u16, RegClass>,
+    /// Set of reftyped vregs that have been defined.
+    ref_regs: HashSet<u16>,
     /// Really a hashmap from rc to HashSet<Reg>.
     regs_by_rc: Vec<HashSet<Reg>>,
     vregs_by_rc: Vec<HashSet<u16>>,
@@ -45,6 +48,11 @@ impl FuzzingEnv {
         !self.regs_by_rc[rc as usize].is_empty()
     }
 
+    /// Returns true whenever a reftyped register may be used.
+    fn can_use_reftyped_reg(&self) -> bool {
+        !self.ref_regs.is_empty()
+    }
+
     /// Returns true whenever a register of the given register class may be defined.
     fn can_def_reg(&self, rc: RegClass) -> bool {
         // If we can use one with the given reg class, then we can redefine it!
@@ -55,6 +63,11 @@ impl FuzzingEnv {
     fn can_def_vreg(&self, rc: RegClass) -> bool {
         !self.vregs_by_rc[rc as usize].is_empty()
             || self.vregs.len() != (self.num_virtual_regs as usize)
+    }
+
+    /// Returns true whenever a reftyped vreg may be defined.
+    fn can_def_reftyped_reg(&self) -> bool {
+        self.ref_regs.len() != (self.num_ref_regs as usize)
     }
 
     fn def_reg(&mut self, rc: RegClass, u: &mut Unstructured) -> Result<Reg> {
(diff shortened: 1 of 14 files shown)


cfallin:
@julian-seward1 I just rebased this and removed the `fuzz_stackmap3.rat` test, since this PR doesn't include a fix; please do include it in your PR with the fix though! (The file is here: https://c1f.net/misc/fuzz_stackmap3.rat)

Also, the last commit in this PR adds some additional `debug!()` logging; if you'd rather that we not pollute the code with this, I'm happy to remove it.

lib/src/checker.rs
```diff
@@ -224,6 +235,37 @@ impl CheckerState {
                     }
                 }
             }
+            &Inst::SSMove {
```
julian-seward1, line 251: `SSMove` feels a bit vague.  Would you consider `SSChangeOwnership` or something like that?

julian-seward1: Urr, I should have really put this comment on the def point for `SSMove`, but you get the idea.

lib/src/bt_main.rs
```diff
@@ -1496,6 +1497,11 @@ pub fn alloc_main<F: Function>(
                 | (Point::Def, Point::Spill, Point::Reload, Point::Use) => {
                     let slot1 = edit_list_move[i_min].slot;
                     let slot2 = edit_list_move[i_max].slot;
+                    let (from_reg, to_reg) = if frag1.last.pt() == Point::Use {
```
julian-seward1, line 1500: AFAICS `(from_reg, to_reg)` are used only in the case `use_checker == true`.  Please move it there, therefore.  We might get lucky in that LLVM can PRE it down to the use point, but I don't want to rely on that, and also, it'd be clearer for the reader.

lib/src/checker.rs
bnjbvr: As a drive-by reader (and feel free to ignore if it doesn't make sense because it's explained somewhere around this code), I don't even understand "SS", could it be expanded too?

lib/src/bt_main.rs
```diff
@@ -1504,6 +1510,17 @@ pub fn alloc_main<F: Function>(
                         iixs_to_nop_out.push(i_min_iix);
                         i_min = i_max + 1;
                         n_edit_list_move_processed += 2;
+                        if use_checker {
```
julian-seward1, line 1513: Ok .. so the cost of this gig, for the production, non-checking case, is one totally predictable conditional branch per coalesced-out move.  That's acceptable.

bin/fuzzing.rs
```diff
@@ -20,8 +20,11 @@ const NUM_BLOCK_INSTS: u8 = 8;
 struct FuzzingEnv {
     num_blocks: u8,
     num_virtual_regs: u16,
+    num_ref_regs: u16, // numbered in vreg space above ordinary vregs.
```
julian-seward1, line 23: nitlet: would you consider `*_reftyped_reg*` instead of `*_ref_reg*` (and consistently below)?  It would be a little easier to follow and also more consistent, eg, look at the current definition of `can_use_reftyped_reg` just below.

bin/test_framework.rs
```diff
@@ -362,6 +362,15 @@ pub enum Inst {
         addr: AM,
         src: Reg,
     },
+    MakeRef {
```
julian-seward1, line 367: Presumably these create a reffy value from a non-reffy value, and vice versa, right?  Is there any checking there (or needed?) that actually ensures this (assertishly, I mean) ?

bin/test_framework.rs
```diff
@@ -1143,19 +1180,25 @@ impl Inst {
...
         match self {
             Value::U32(x) => match other {
                 Value::U32(y) => x == y,
-                Value::F32(_) => false,
+                _ => false,
             },
 
             Value::F32(x) => match other {
-                Value::U32(_) => false,
                 Value::F32(y) => (x != x && y != y) || x == y,
```
julian-seward1, line 1194: (I know this is pre-existing, but ..) doesn't Rust have `isNaN` or `!isNormal` or something like that, by which we could avoid the gruesomely hacky feeling that comes from comparing a FP value with itself?

lib/src/bt_spillslot_allocator.rs
```diff
@@ -11,6 +11,8 @@ use crate::data_structures::{
 use crate::union_find::UnionFindEquivClasses;
 use crate::Function;
 
+use log::debug;
```
julian-seward1, line 14: Would you be happy to remove the debug printing additions here?  I would prefer that.  For the perhaps lame reason that it will make the existing `RUST_LOG=debug` debug printing for RA/BT even more verbose, and IIUC this is here really only to track down initial implementation bugs.  Yes .. I know .. the RA/BT debug printing should really be done with `trace!` and not `debug!`.

julian-seward1:
Duh.  I meant to write more.

OK to land but please action at a minimum, the comment referring to LLVM and PRE, and the one regarding removal of debug printing.

More broadly -- thanks for fixing this.  It's obviously valuable.  Also, thank you for the sometimes-epic commit messages.  They really help.  I have some comments/clarifications for that of the first patch; not saying you should incorporate them all, but do take a look.  They follow.

-----------

> for example, a move from one virtual register to another, when both
> virtual registers live in the same spillslot (a "V-V move").

A V-V move doesn't *necessarily* mean both vregs get assigned the same spillslot.  That would
require that:

(1) they are coalescable; meaning the move is a last-use of the src and a first-def of the dst.

(2) even with (1), there's no *guarantee* that SpillSlotAllocator will give them the same spill
slot.  (It will try hard to but doesn't guarantee that).

In the sources, I have tended to call V-V moves that satisfy (1) "V-V boundary moves".

> that it also loses information, because we may have two different
> programs that store a different vreg in a slot when arriving at a

"programs" doesn't make sense at that point.  Did you perhaps mean "instructions" ?

> (the checker should be simple and straightforward to maximize the
> confidence we have in it).

Agreed.

julian-seward1:
> "programs" doesn't make sense at that point. Did you perhaps mean "instructions" ?

Or maybe you meant "basic blocks"?

lib/src/checker.rs
cfallin: Yes, `ChangeSpillSlotOwnership` now -- much clearer, thanks.

lib/src/bt_main.rs
cfallin: Done!

lib/src/bt_main.rs
cfallin: Indeed, I think this is the best we can do (unless we lift the branch to entry and clone the rest of the code somehow).

bin/fuzzing.rs
cfallin: Done!

bin/test_framework.rs
cfallin: It's implicitly ensured by the fuzzing test generation at least, by keeping separate pools of reffy and non-reffy registers and pulling from the appropriate ones for each instruction. I started down the path of adding reffyness to the typechecking here but it operates on the basis of register class rather than a separate type and I didn't want to refactor the whole thing; this seems "good enough" for the purposes of fuzzing, I think.

bin/test_framework.rs
cfallin: TIL `.is_nan()` -- done.

lib/src/bt_spillslot_allocator.rs
cfallin: Yup, stripped out that commit.

cfallin:
Thanks! Addressed comments, and edited the first commit message. (I think I had meant "program point" but "from different paths" is more correct.)

PR #86, opened 2020-07-16
In the spill slot allocator, set `RangeFragAndRefness::is_ref` to `tr…

julian-seward1:
…ue` only when we know that

a reftyped `VirtualRange` is going to be spilled.  This fixes a conceptual error in stackmap
generation found by Chris via fuzzing.  The error was that `RangeFragAndRefness::is_ref` was
previously set to `true` too early.  Specifically, spill slots for all `VirtualRanges` of a
virtual range equivalence class are allocated at the point where the first `VirtualRange` in the
eclass was spilled.  And the `RangeFragAndRefness::is_ref` of all `RangeFrag`s owned by the
eclass was set to `true` if the `VirtualRange` was reftyped.

From the point of view of "does this spill slot fragment carry a reftyped thing, or not?",
that's exactly correct.  However, it ignores the fact that other `VirtualRange`s in the same
eclass ultimately might not be spilled.  Hence the effect is to potentially mark some of those
spill slot sections as reftyped even though they will never be used.  Consequently the client
could end up passing such slot sections to its GC for scavenging etc even though they contain
undefined values.  This could cause crashing or worse.

The fix is (conceptually) very simple: don't set `RangeFragAndRefness::is_ref` to `true` when
initially allocating spill slots.  Instead, set it (them) later in the process, when spilling a
`VirtualRange` that is reftyped.  A new method,
`SpillSlotAllocator::notify_spillage_of_reftyped_vlr`, has been added for that purpose.

This was found by Chris' regalloc-result checker and fuzzing.  That is fortunate, since it would
have otherwise caused an obscure GC crash which would have been hard to track down.

### DIFF
--- a/lib/src/bt_spillslot_allocator.rs
+++ b/lib/src/bt_spillslot_allocator.rs
@@ -168,15 +168,11 @@ fn ssal_is_add_possible(tree: &AVLTree<RangeFragAndRefness>, frags: &SortedRange
 }
 
 // HELPER FUNCTION
-// Try to add all of `frags` to `tree`.  Return `true` if possible, `false` if
-// not possible.  If `false` is returned, `tree` is unchanged (this is
-// important).  This routine relies on the fact that SortedFrags is
-// non-overlapping.
-fn ssal_add_if_possible(
-    tree: &mut AVLTree<RangeFragAndRefness>,
-    frags: &SortedRangeFrags,
-    frags_are_ref: bool,
-) -> bool {
+// Try to add all of `frags` to `tree`.  Return `true` if possible, `false` if not possible.  If
+// `false` is returned, `tree` is unchanged (this is important).  This routine relies on the
+// fact that SortedFrags is non-overlapping.  They are initially all marked as non-reffy.  That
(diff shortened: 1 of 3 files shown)


lib/src/bt_spillslot_allocator.rs
```diff
@@ -195,6 +191,27 @@ fn ssal_add_if_possible(
     true
 }
 
+// HELPER FUNCTION
+// Let `frags` be the RangeFrags for some VirtualRange, that have already been allocated in
+// `tree`.  Mark each such RangeFrag as reffy.
+fn ssal_mark_frags_as_reftyped(tree: &mut AVLTree<RangeFragAndRefness>, frags: &SortedRangeFrags) {
+    for frag in &frags.frags {
+        // Be paranoid.  (1) `frag` must already exist in `tree`.  (2) it must not be marked as
+        // reffy.
```
cfallin, line 200: This is true because we only ever mark a vlr as reffy once, because once we decide to spill, we don't un-spill, is that right? It would be good to document this invariant and hence why the find-and-replace works in a comment here.

julian-seward1: Good point.  I updated the comment, in some detail.

PR #90, opened 2020-08-09
Rewrite `AVLTree::insert_wrk` to be non-recursive.  Profiling shows this

julian-seward1:
function to be hot in the case where the function to be allocated contains
very many function calls, since then there will be a huge number of
minimum-length real reg live ranges to be inserted into commitment trees.

The rewrite here was created by mechanical transformation of the recursive
original through many intermediate steps, all of which functioned correctly.

The supporting functions `leftgrown` and `rightgrown` have been split into hot
and cold parts for better inlining.

For the `joey_big` test case, this reduces compiler insn count by 1.1%;
smaller functions get smaller wins.  For some of the pathological test cases
in Wasmtime PR911, there is a wallclock time win of up to 2%.

From those pathological test cases it is also clear that, at least for
themselves, there is a much larger win on the table here: since the tens of
thousands of RangeFrags to be inserted into each tree are in order, it's
wasteful to insert them one by one.  Instead the sorted vector of them, that
is pushed into the tree, could be transformed in linear time into an almost
perfectly balanced binary tree.  Whether this is worth it for any realistic
inputs seems doubtful, however.

### DIFF
--- a/lib/src/avl_tree.rs
+++ b/lib/src/avl_tree.rs
@@ -3,6 +3,7 @@
 //! AVL tree internals are public, so that backtracking.rs can do custom
 //! traversals of the tree as it wishes.
 
+use smallvec::SmallVec;
 use std::cmp::Ordering;
 
 //=============================================================================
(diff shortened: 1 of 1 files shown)


lib/src/avl_tree.rs
```diff
@@ -310,67 +316,119 @@ impl<T: Clone + PartialOrd> AVLTree<T> {
...
+        // In the first phase, walk down the tree to find the place where the new node should be
+        // inserted.  This loop is cloned so as to allow the test on `mb_cmp` to be done just
+        // once.
+        match mb_cmp {
+            None => {
+                while root != AVL_NULL {
+                    let cmp_loc_right = &self.pool[root as usize];
+                    let cmp_arg_right: T = cmp_loc_right.item.clone();
+                    let cmp_arg_left: T = item.clone();
+                    debug_assert!((root & 0x8000_0000) == 0);
```
cfallin, line 347: I'd prefer to pull out the bitpacking into inlined helpers, just to centralize the scheme's definition -- perhaps `stack_entry_is_left()`, `stack_entry_node_index()`, and `make_stack_entry()` or something like that.

PR #95, opened 2020-08-17
Use `SmallVec` inside `RegToRangesMaps::rreg_to_rlrs_map`.  This redu…

julian-seward1:
…ces the malloc count by

almost 5% for CL/x64/newBE compiling ZenGarden.wasm, and appears to be performance neutral on
other test cases.

### DIFF
--- a/lib/src/data_structures.rs
+++ b/lib/src/data_structures.rs
@@ -2107,9 +2107,10 @@ pub struct RegToRangesMaps {
     // This maps RealReg indices to the set of RealRangeIxs for that RealReg.  Valid indices are
     // real register indices for all non-sanitised real regs; that is,
     // 0 .. RealRegUniverse::allocable, for ".." having the Rust meaning.  The Vecs of
-    // RealRangeIxs are duplicate-free.  They are Vec rather than SmallVec because they are often
-    // large, so SmallVec would just be a disadvantage here.
-    pub rreg_to_rlrs_map: Vec</*real reg ix, */ Vec<RealRangeIx>>,
+    // RealRangeIxs are duplicate-free.  The SmallVec capacity of 6 was chosen after quite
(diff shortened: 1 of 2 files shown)


lib/src/data_structures.rs
```diff
@@ -2107,9 +2107,10 @@ pub struct RegToRangesMaps {
     // This maps RealReg indices to the set of RealRangeIxs for that RealReg.  Valid indices are
     // real register indices for all non-sanitised real regs; that is,
     // 0 .. RealRegUniverse::allocable, for ".." having the Rust meaning.  The Vecs of
-    // RealRangeIxs are duplicate-free.  They are Vec rather than SmallVec because they are often
-    // large, so SmallVec would just be a disadvantage here.
-    pub rreg_to_rlrs_map: Vec</*real reg ix, */ Vec<RealRangeIx>>,
+    // RealRangeIxs are duplicate-free.  The SmallVec capacity of 6 was chosen after quite
```
bnjbvr, line 2110: (light suggestion) since you're mentioning profiling, could you add a note at which date profiling was done? it's likely that external things can have an impact on profiles (so, all things being equal in regalloc, it may slow down or get faster if changing the numbers because of external effects in e.g. isel), and this also gives a reasonable hint for when profiling has been done last (thus give a guess of when it's too early to re-profile or not).

julian-seward1: Ah, that is an excellent suggestion.  I'll also put in a line summarising *what* was profiled.

Amanieu:
If memory allocation occupies a significant portion of the runtime then you might be interested in using a bump-pointer allocator such as [bumpalo](https://github.com/fitzgen/bumpalo). Allocating memory becomes just a handful of instructions and all memory allocations can be bulk-freed after register allocation is finished.

PR #99, opened 2020-10-05
Do not use an enum for SpillSlot, and simplify Writable constraints + import pretty printing

bnjbvr:
No change in functionality here.

This will require changes in Cranelift too (so Cranelift gets rid of its own ShowWithRRU trait), but fortunately these changes are non-breaking, since the two traits can coexist.


cfallin (approved):
This is a nice cleanup, thanks!

PR #100, opened 2020-10-06
Add `is_included_in_clobbers` to public API.

cfallin:
This allows the client to request that certain instructions' register
writes be excluded from the returned clobbered-register set. This may be
useful when the client can handle such writes in other ways: for
example, by reasoning directly about ABIs. (One common case is that if a
caller and callee have the same ABI, then the call instruction in the
caller need not be included in the clobber set that is used to generate
the caller's clobber-saves, because any register that the callee
clobbers is also allowed to be clobbered by the caller.)

See also bytecodealliance/regalloc.rs#2228 and bytecodealliance/regalloc.rs#2254.

### DIFF
--- a/lib/src/bt_main.rs
+++ b/lib/src/bt_main.rs
@@ -1792,7 +1792,9 @@ pub fn alloc_main<F: Function>(
     let mut reg_vecs = RegVecs::new(/*sanitized=*/ false);
     let mut dummy_bounds = RegVecBounds::new();
     for insn in &final_insns {
-        add_raw_reg_vecs_for_insn::<F>(insn, &mut reg_vecs, &mut dummy_bounds);
+        if func.is_included_in_clobbers(insn) {
+            add_raw_reg_vecs_for_insn::<F>(insn, &mut reg_vecs, &mut dummy_bounds);
+        }
     }
     for reg in reg_vecs.defs.iter().chain(reg_vecs.mods.iter()) {
         assert!(reg.is_real());
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -224,6 +224,25 @@ pub trait Function {
     /// Determine whether an instruction is a return instruction.
     fn is_ret(&self, insn: InstIx) -> bool;
 
+    /// Determine whether an instruction should be considered while computing
+    /// the set of registers that need to be saved/restored in the function's
+    /// prologue/epilogue, that is, the registers returned in
+    /// `clobbered_registers` in `RegAllocResult`.  computation. Only
+    /// instructions for which this function returns `true` will be used to
+    /// compute that set.
+    ///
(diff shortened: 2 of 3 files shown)


lib/src/bt_main.rs
```diff
@@ -1792,7 +1792,9 @@ pub fn alloc_main<F: Function>(
     let mut reg_vecs = RegVecs::new(/*sanitized=*/ false);
     let mut dummy_bounds = RegVecBounds::new();
     for insn in &final_insns {
-        add_raw_reg_vecs_for_insn::<F>(insn, &mut reg_vecs, &mut dummy_bounds);
+        if func.is_included_in_clobbers(insn) {
```
julian-seward1, line 1795: Meh.  "Greetings, Earthling; take me to your Branch Predictor" .. kinda thing.  But I see it's unavoidable.  Are we at least reasonably confident that the `is_included_in_clobber` call will be inlined?  That's possibly important.  For sure we don't want to wind up with this being a virtual method call.

lib/src/lib.rs
```diff
@@ -220,6 +220,23 @@ pub trait Function {
     /// Determine whether an instruction is a return instruction.
     fn is_ret(&self, insn: InstIx) -> bool;
 
+    /// Determine whether an instruction should be included in clobber-set
+    /// computation. Only instructions for which this function returns `true`
+    /// will have their defs and mods included in `clobbered_registers` in the
+    /// `RegAllocResult`.
+    ///
```
julian-seward1, line 227: Given that the RA interface is already semantically complex, I'd prefer to make the wording a bit more specific here, in order to make clear that this isn't related to the live range computation that the allocator has to do, that involves `fn get_regs`.  Maybe something like:

Determine whether an instruction should be considered whilst computing the set of registers that need to saved/restored in the function's prologue/epilogue, that is, `clobbered_registers` in the `RegAllocResult`.  Only instructions for which this function returns `true` will be used to compute that set.

julian-seward1 (approved):
LGTM, but please update comment as described (or similarly).

lib/src/bt_main.rs
cfallin: Indeed, I just verified that it's inlined in a release build .(Steps: built `wasmtime` on aarch64; disassembled; grepped for `is_included_in_clobbers`; no results. Verified that in a debug build, the symbol is present, so I'm grepping for the right thing and opts make it disappear.)

If we wanted to force this to be extra-super-sure, even in debug builds, it seems we can't put a directive on the *trait* (thread about this [here](https://users.rust-lang.org/t/trait-method-inlining/16817)), but any client could add a directive on their impl.

lib/src/lib.rs
cfallin: Done, thanks!

PR #102, opened 2020-10-20
`get_range_frags_for_block`: improve cost handling for 'modifies' ref…

julian-seward1:
…erences to a reg.

One of the tasks of `fn get_range_frags_for_block` is to count the number of times each virtual
reg is mentioned, so as to give a basis for computation of spill costs.  It adds one for a
'use' or 'def' reference, which is reasonable.  But it also currently adds one for a 'mod'
reference, which is a bit misleading, given that if we spill a live range containing a 'mod'
reference, we'll have to generate a reload before the referencing insn and a spill afterwards.
So really we should increment the menioned-count by two, not one.  This patch makes that change.

### DIFF
--- a/lib/src/analysis_data_flow.rs
+++ b/lib/src/analysis_data_flow.rs
@@ -1004,8 +1004,11 @@ fn get_range_frags_for_block<F: Function>(
                 // not listed in `livein`, since otherwise `state` would have an entry for it.
                 None => panic!("get_range_frags_for_block: fail #2"),
                 Some(ref mut pf) => {
-                    // This the first or subsequent modify after a write.
-                    pf.num_mentions = plus1(pf.num_mentions);
+                    // This the first or subsequent modify after a write.  Add two to the
+                    // mentions count, as that reflects the implied spill cost increment more
+                    // accurately than just adding one: if we spill the live range in which this
+                    // ends up, we'll generate both a reload and a spill instruction.
+                    pf.num_mentions = plus1(plus1(pf.num_mentions));
                     let new_last = InstPoint::new_def(iix);
                     debug_assert!(pf.last <= new_last);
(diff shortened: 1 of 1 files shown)


lib/src/analysis_data_flow.rs
```diff
@@ -1004,8 +1004,11 @@ fn get_range_frags_for_block<F: Function>(
                 // not listed in `livein`, since otherwise `state` would have an entry for it.
                 None => panic!("get_range_frags_for_block: fail #2"),
                 Some(ref mut pf) => {
-                    // This the first or subsequent modify after a write.
-                    pf.num_mentions = plus1(pf.num_mentions);
+                    // This the first or subsequent modify after a write.  Add two to the
+                    // mentions count, as that reflects the implied spill cost increment more
+                    // accurately than just adding one: if we spill the LR in which this ends
```
bnjbvr, line 1009: nit: can you expand live range, please?

lib/src/analysis_data_flow.rs
```diff
@@ -1004,8 +1004,11 @@ fn get_range_frags_for_block<F: Function>(
...
+                    // up, we'll generate both a reload and a spill instruction.
+                    pf.num_mentions = plus1(plus1(pf.num_mentions));
```
bnjbvr, line 1011: (An idea for a future refactoring) It seems that instead of the plus1 function, we could use `saturating_add` on the field itself, avoiding the weird double plus1 call.

bnjbvr (approved):
Makes sense! Note there's a typo in the commit message: `menioned` should be `mentioned` instead.

To wit: since LSRA's forked this code, and LSRA keeps track of precise mentions, there's nothing to change there.

julian-seward1:
Thanks for the review!

PR #105, opened 2020-11-19
Checker: make sure that ref values live over a safepoint must be reloaded

bnjbvr:
Also adds enough testing support so that we can actually run a small
test program with reftypes, as well as a unit test.


cfallin (approved):
LGTM -- thanks very much for finding this gap!

PR #107, opened 2020-11-19
Tweak the crate/workspace setup so that fuzz binaries are built in root `target/` directory.

cfallin:
Because the fuzzing crate was previously in its own Cargo workspace, its
build process placed its binaries in `fuzz/target/...`. Other
infrastructure (such as OSS-Fuzz) doesn't expect this. This PR instead
makes the fuzzing crate an ordinary part of the main workspace, and
tweaks the build-option overrides to work with this new setup. Followup
to #106.

Sorry for the tedium of little tweaks here!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,8 @@ publish = false
 
 [workspace]
 members = [
-    "lib"
+    "lib",
+    "fuzz"
 ]
 
 [[bin]]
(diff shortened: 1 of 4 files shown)


cfallin:
Urgh, something's wrong with the build-option overrides; standby...

alexcrichton (approved):

alexcrichton:
Oh this is where if the fuzzers are a member of the main workspace then `cargo test --all` will try to test the fuzzers which isn't what you want.

We fixed this in Wasmtime with some [extra directives](https://github.com/bytecodealliance/wasmtime/blob/acbe85fa95312f0bf4a35ea245185e0f0964dc4a/fuzz/Cargo.toml#L61-L62) which, while ugly, at least get the job done.

PR #108, opened 2020-11-20
Re-enable checker for LSRA

bnjbvr:
I've been fortunate; there were no new LSRA failures since the checker had been deactivated.

This also adds two small things:

- add a check that the checker is eventually run, otherwise it's useless. We could remove this part, since it might fiddle with fuzzing in surprising ways, e.g. if fuzzing finds a bug somewhere between the creation of `CheckerContext` and `CheckerContext::run()`, it may panic while it panics because of this change.
- don't mix "having no reftype_start" with "all vregs are reftypes". Use a real `Option` to denote the presence or absence of this configuration line. Also renders the special annotation when running from the fuzzer: this should help with extracting test cases from the fuzzer output.

### DIFF
--- a/bin/fuzzing.rs
+++ b/bin/fuzzing.rs
@@ -431,7 +431,7 @@ impl Arbitrary for Func {
             name: "funk".to_string(),
             entry,
             num_virtual_regs: (num_virtual_regs + num_reftyped_regs) as u32,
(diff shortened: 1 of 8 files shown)


cfallin (approved):
Looks good!

PR #109, opened 2020-11-23
Two additional LSRA fixes

bnjbvr:
This makes LSRA pass the wasmtime tests. See commit messages for further descriptions.


cfallin (approved):
Nice debugging! Very exciting that this now passes tests.

PR #103, opened 2020-11-17
Refactor analysis

bnjbvr:
Small refactorings I saw while wanting to work on linear scan. This is mostly using more idiomatic Rust in a few places, using doc comments, wrapping comment lines at 100 chars, simplifying a few patterns, trying to encapsulate a few structs a bit better and enhance their ergonomics (e.g. `MoveInfo`). No changes in functionality here. Commit messages contain more details about the nature of each change.

@julian-seward1 and/or @cfallin, PTAL. Thanks!

PR #110, opened 2020-11-23
Support reference types in linear scan

bnjbvr:
This builds on top of #103 and #109 to add support for reference types to linear scan.

The implementation is slightly different from the one used for backtracking. Backtracking uses a post-processing approach: during analysis, ranges are marked as reffy (i.e. containing reference types); after register allocation, a spill/reload sequence is generated safepoint instruction contained in a reffy range. Since linear scan has support for splitting live ranges, it is able to integrate reference type processing during register assignment: when an interval (~range) is assigned a real register, if it contains a safepoint instruction, then the interval is split and spilled at the safepoint instruction. A single reffy interval `[ start .. safepoint .. end ]` becomes multiple intervals:

- `[ start .. (prev mention .. safepoint)[` allocated to the register; the split point may be in any instruction between just after the previous mention, and just before safepoint,
- `[safepoint .. next mention]`, the interval stays stack allocated until the next real mention of the virtual register
- `[next mention .. end]` is an unallocated interval pushed back into the allocation queue

This is the general scheme. If the safepoint instruction redefines the register (the register is a use and a def of the call), there's no need to spill and reload it across the instruction.
 
This doesn't handle reffy real ranges, since neither Cranelift or the testing framework effectively put reference types in fixed real registers. The LSRA analysis asserts this right now. In the future, we could consider relaxing the regalloc invariants so that only virtual registers can be reffy, and reffy moves involve both a reffy source vreg and reffy target vreg. If so, we could get rid of the reftypes analysis and all the required bookkeeping data structures LSRA has for this.

### DIFF
--- a/fuzz/fuzz_targets/lsra.rs
+++ b/fuzz/fuzz_targets/lsra.rs
@@ -4,34 +4,57 @@ use libfuzzer_sys::fuzz_target;
 use minira::{self, test_framework as ir};
 use regalloc;
 
+static mut COUNTER_GEN: usize = 0;
+static mut COUNTER_OK: usize = 0;
+
 fuzz_target!(|func: ir::Func| {
+    let (num_gen, num_ok) = unsafe {
+        COUNTER_GEN += 1;
+        (COUNTER_GEN, COUNTER_OK)
+    };
+
+    println!(
+        "=== status: #ok/#total: {}/{} == {} ",
+        num_ok,
+        num_gen,
+        100.0 * (num_ok as f64) / (num_gen as f64)
+    );
+
     let mut func = func;
+    let original_func = func.clone();
 
     // Add one register to be used as a scratch.
     let num_regs = minira::fuzzing::NUM_REAL_REGS_PER_RC as usize + 1;
     let reg_universe = ir::make_universe(num_regs, num_regs);
 
-    let mut rendered = String::new();
-    func.render("func:", &mut rendered).unwrap();
-    println!("{}", rendered);
-
+    let sri = func.get_stackmap_request();
     let result = match regalloc::allocate_registers_with_opts(
         &mut func,
         &reg_universe,
-        None,
+        sri.as_ref(),
         regalloc::Options {
             run_checker: true,
             algorithm: regalloc::Algorithm::LinearScan(Default::default()),
         },
     ) {
-        Ok(result) => result,
+        Ok(result) => {
+            unsafe {
+                COUNTER_OK += 1;
+            }
+            result
+        }
         Err(err) => {
             if let regalloc::RegAllocError::RegChecker(_) = &err {
+                let mut rendered = String::new();
+                original_func.render("func:", &mut rendered).unwrap();
+                println!("{}", rendered);
+
                 panic!(format!(
                     "fuzz_lsra_differential.rs: checker error: {:?}",
                     err
                 ));
             }
+
             println!("allocation error: {}", err);
             return;
         }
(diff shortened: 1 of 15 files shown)


fuzz/fuzz_targets/lsra.rs
```diff
@@ -4,31 +4,54 @@ use libfuzzer_sys::fuzz_target;
...
-    func.render("func:", &mut rendered).unwrap();
-    println!("{}", rendered);
-
+    let sri = func.get_stackmap_request();
     let result = match regalloc::allocate_registers(
         &mut func,
         &reg_universe,
-        None,
+        sri.as_ref(),
         regalloc::AlgorithmWithDefaults::LinearScan,
```
cfallin, line 35: I may be missing something, but I think that `allocate_registers` calls `allocate_registers_with_opts` using the `Default` impl of `regalloc::Options` as a base (then filling in this algorithm), which results in `use_checker` defaulting to `false` -- so is the checker running here? I recall you mentioning a number of checker issues with LSRA + reftypes so I'm sure I'm just missing the plumbing somewhere...

lib/src/analysis_main.rs
```diff
@@ -48,6 +51,8 @@ pub enum AnalysisError {
     /// Implementation limits exceeded.  The incoming function is too big.  It
     /// may contain at most 1 million basic blocks and 16 million instructions.
     ImplementationLimitsExceeded,
+
+    LsraReffyRealReg(RealReg),
```
cfallin, line 55: How does this interact with reffy args/return values to a callsite that is a safepoint? Or is this accepted because args are dead after the use-point and before the "actual safepoint" at the call action itself (and likewise for returns)? I think it would be worth some doc-commentage here to clarify exactly what situations are and are not allowed.

lib/src/analysis_main.rs
```diff
@@ -75,6 +80,9 @@ impl ToString for AnalysisError {
                 "implementation limits exceeded (more than 1 million blocks or 16 million insns)"
                     .to_string()
             }
+            AnalysisError::LsraReffyRealReg(reg) => {
+                format!("an instruction has put a reference type in thel real register {:?}, which is disallowed in linear scan" ,reg)
```
cfallin, line 84: s/thel/the/

lib/src/analysis_reftypes.rs
```diff
@@ -6,10 +6,83 @@ use crate::data_structures::{
...
+
+    /// Add all the ranges associated to this vreg into the set of reftyped ranges.
+    fn insert_reffy_ranges(&self, vreg: VirtualReg, set: &mut SparseSet<Self::RangeId>);
+
+    /// Mark a given RangeId as being reffy.
+    fn mark_reffy(&mut self, range_id: &Self::RangeId);
+}
+
+/// Implementation of the reftype analysis for the backtracking algorithm.
+struct BacktrackingReftypeAnalysis<'a> {
```
cfallin, line 31: Perhaps we could put this in the BT allocator instead (and keep this whole file abstracted)?

cfallin (approved):
Very nice!

The abstracted analysis code is much cleaner; strong r+ on that part.

I am less familiar with the LSRA part of this PR, but on a skim it all looks reasonable, and if it passes a good round of checker-fuzzing then it's good enough for me!

A few nits below but otherwise happy to see this merged.

fuzz/fuzz_targets/lsra.rs
bnjbvr: Now addressed as part of #114.

lib/src/analysis_main.rs
bnjbvr: Actually the analysis error was meaningless: I thought real regs could in theory appear in stack maps, but they don't need to, in our model. I'll remove this error and associated code.

cfallin: Sounds good!

cfallin (approved):

PR #111, opened 2021-01-03
Mark the crate as `#![no_std]`

tomaka:
This marks `regalloc` as `#![no_std]`, allowing its usage on platforms that don't have a standard library.

The PR was mostly mechanical, except:

- I use `hashbrown` to provide `HashMap`/`HashSet`. This is the same crate that [the `std` itself uses](https://github.com/rust-lang/rust/blob/18cb4ad3b9440b3ff2ed16976f56889b23811e13/library/std/Cargo.toml#L23).
- I replaced two `println!`s with `info!` (from the `log` crate).
- There are a few usages of `std::env` which I can't replace with an equivalent (see below).

The code in `src/linear_scan/mod.rs` uses `std::env`, apparently for debugging purposes.
At the moment I've just commented out the code, and this PR is marked as draft.

I'd suggest adding something like a `debug-env-override` feature to the crate, which, when enabled, disables `#![no_std]` and makes it use `std::env` as is currently the case.
That's just a suggestion, so let me know if that's suitable or if there's a better solution.

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -8,7 +8,8 @@ description = "Modular register allocation algorithms"
 repository = "https://github.com/bytecodealliance/regalloc.rs"
 
 [dependencies]
-rustc-hash = "1.0.1"
+hashbrown = { version = "0.9.1", default-features = false }
+rustc-hash = { version = "1.0.1", default-features = false }
 log = { version = "0.4.8", default-features = false }
 smallvec = "1.0.0"
 serde = { version = "1.0.94", features = ["derive"], optional = true }
@@ -18,4 +19,4 @@ default = []
 fuzzing = []
 
 # For dependent crates that want to serialize some parts of regalloc.
-enable-serde = ["serde"]
+enable-serde = ["serde", "hashbrown/serde"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ path = "./bin/minira_lib.rs"
 [dependencies]
 regalloc =  { path = "./lib", features = ["fuzzing", "enable-serde"] }
 arbitrary = { version = "0.4.0", features = ["derive"]}
-rustc-hash = "1.0.1"
+rustc-hash = { version = "1.0.1", default-features = false }
 log = { version = "0.4.8", default-features = false }
 pretty_env_logger = "0.4.0"
 clap = "2.33.0"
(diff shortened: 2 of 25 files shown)


lib/Cargo.toml
```diff
@@ -18,4 +19,4 @@ default = []
 fuzzing = []
 
 # For dependent crates that want to serialize some parts of regalloc.
-enable-serde = ["serde"]
+enable-serde = ["serde", "hashbrown/serde"]
```
bjorn3, line 22: This will forcefully enable hashbrown when using `enable-serde`.

tomaka: `hashbrown` isn't an optional dependency, or am I misunderstanding?

bjorn3: Oops, read `default-features = false` as `optional = true` for some reason. :smile:

bnjbvr:
Thanks! I'll take a closer look to all the changes later, only two small comments right now:

- re: the LSRA code: a new feature sounds good (we could even call it `lsra-tweaks` or so), as long as we enable it by default in the toy app (top-level Cargo). Can you add this please?
- I'm not sure what's the preferred way in the Rust community, but would it make sense to have a single file with all the type aliases for hashmaps etc., so that other files can import those, and use that instead of redefining the hashmaps every time?

tomaka:
I've added the `lsra-tweaks` feature as requested.

> I'm not sure what's the preferred way in the Rust community, but would it make sense to have a single file with all the type aliases for hashmaps etc., so that other files can import those, and use that instead of redefining the hashmaps every time?

In theory one is encouraged to tweak the hashing algorithm depending on the nature of the key. Having some sort of "global type alias" goes in contradiction with this.

bnjbvr:
Hi! Sorry for the long response time. This crate will be deprecated in favor of [regalloc2](https://github.com/bytecodealliance/regalloc2) in the future, at least for usage in Cranelift/Wasmtime, so subsequent additions and improvements should happen there. I would suggest starting with a discussion in issues first, to know whether the maintainers want to support no-std in the long term. In the meanwhile, closing this.

tomaka:
> I would suggest starting with a discussion in issues first, to know whether the maintainers want to support no-std in the long term.

I've unfortunately received negative feedback about this in cranelift/wasmtime, for code style reasons that I don't really understand.

PR #114, opened 2021-01-14
LSRA bug fixes

bnjbvr:
This re-enables the checker for fuzzing, and fixes two fuzz bugs it did find. The commit messages explain what the issues are in detail. With these fixes in, I could get this branch + #110 fuzzed for half an hour and zarro boogs found.

### DIFF
--- a/lib/src/linear_scan/assign_registers.rs
+++ b/lib/src/linear_scan/assign_registers.rs
@@ -1,6 +1,6 @@
 use super::{
-    last_use, next_use, IntId, Intervals, Mention, MentionMap, OptimalSplitStrategy, RegUses,
-    Statistics, VirtualInterval,
+    analysis::BlockPos, last_use, next_use, IntId, Intervals, Mention, MentionMap,
+    OptimalSplitStrategy, RegUses, Statistics, VirtualInterval,
 };
 use crate::{
     data_structures::{InstPoint, Point, RegVecsAndBounds},
@@ -1132,6 +1132,7 @@ fn split<F: Function>(state: &mut State<F>, id: IntId, at_pos: InstPoint) -> Int
     let vreg = int.vreg;
     let ancestor = int.ancestor;
 
+    // Split the register mentions.
     let parent_mentions = state.intervals.get_mut(id).mentions_mut();
     let index = parent_mentions.binary_search_by(|mention| {
         // The comparator function returns the position of the argument compared to the target.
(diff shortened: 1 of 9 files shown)


lib/src/linear_scan/resolve_moves.rs
```diff
@@ -337,14 +306,18 @@ fn resolve_moves_across_blocks<F: Function>(
...
         //
-        // We assume critical edges have been split, so
-        // if the current block has more than one successor, then its successors
-        // have at most one predecessor.
-        let cur_has_one_succ = successors.len() == 1;
+        // If there's more than one successor to the current block, then inserting the fix-up moves
+        // in the current block may impact all the other successors: in this case, we should insert
+        // the instruction at the end of the current block. If there's only successor, or the
+        // instruction has register mentions (see also: `LsraBranchWithUses`), then we can insert
+        // the fixup move at the beginning of the successor safely.
```
cfallin, line 317: Would it make sense to use the heuristic "always insert in successor block, unless it has multiple preds" instead, to avoid the need to count register mentions in the last instruction? I think that should give correct results given the LSRA crit-edge invariant.

cfallin (approved):
Looks good overall! I don't claim to fully understand the block-edge / overlapping-intervals details in the last commit; perhaps Julian could double-check that bit (or I defer to your understanding + confirmation by fuzzing). Otherwise, I'm really happy that LSRA seems to be fully usable now!

lib/src/linear_scan/assign_registers.rs
```diff
@@ -1179,24 +1179,6 @@ fn split<F: Function>(state: &mut State<F>, id: IntId, at_pos: InstPoint) -> Int
     }
     parent_mentions.truncate(index);
 
-    // In the situation where we split at the def point of an instruction, and the mention set
-    // contains the use point, we need to refine the sets:
-    // - the parent must still contain the use point (and the modified point if present)
-    // - the child must only contain the def point (and the modified point if present).
-    // Note that if we split at the use point of an instruction, and the mention set contains the
-    // def point, it is fine: we're not splitting between the two of them.
-    if may_need_fixup && at_pos.pt() == Point::Def && child_mentions.first().unwrap().1.is_use() {
```
bjorn3, line 1188: Maybe assert that the fixup condition is false?

bnjbvr: Something I've considered, so I tried it: I added a panic in this block and ran fuzzing for long enough and it didn't trigger. Plus it's provable that this can't happen: `child_mentions.first()` would a def, if `at_pos` was a def.

lib/src/linear_scan/resolve_moves.rs
bnjbvr: Ah, good idea! Note to self: it's not available in `Function`, but it is available in `CFGInfo` computed during analysis.

cfallin (approved):
r+'ing again after discussing -- the block start/end changes look good.

PR #121, opened 2021-05-13
Add README to regalloc crate

olivierlemasle:
Add README to regalloc crate

I've copied the README from the root of the repo. Not sure if it's the best description of the crate...


cfallin (approved):
Thanks!

And, indeed, this README is quite out-of-date; the regalloc is now production-ready, and used by default in Cranelift (hence Wasmtime and Lucet). We should update this, but we can merge this first so we don't block your packaging efforts!

PR #123, opened 2021-05-27
Demote all logging statements from `info!` to `debug!`

pepyakin:
We embed wasmtime in [Substrate](https://github.com/paritytech/substrate). We are looking to upgrade to 0.27 which seems to be using the new backend, even for x64 (🎉  btw!). 

We've noticed a problem though. In Substrate we have a logger enabled and by default it uses `info!` to communicate information mostly used for node operators. After the upgrade though there are logging from regalloc. 

I think we can mute the `regalloc` on our side in the logger configuration. However, from what I understand, libraries should strive away from using `info!`. Instead, they should restrict themselves to be under `debug!` and `trace!`. 

Therefore, I figured that I would send a naive PR here to demote all the `info!` to `debug!`. I understand that this may not be appropriate: let me know if I can change something in this PR or we should just silence regalloc on our part.

### DIFF
--- a/lib/src/linear_scan/assign_registers.rs
+++ b/lib/src/linear_scan/assign_registers.rs
@@ -8,7 +8,7 @@ use crate::{
     VirtualReg, NUM_REG_CLASSES,
 };
 
-use log::{debug, info, log_enabled, trace, Level};
+use log::{trace, log_enabled, Level};
 use rustc_hash::FxHashMap as HashMap;
 use smallvec::SmallVec;
 use std::collections::BinaryHeap;
@@ -204,7 +204,7 @@ pub(crate) fn run<F: Function>(
     let mut prev_start = InstPoint::min_value();
 
     while let Some(id) = state.next_unhandled() {
-        info!("main loop: allocating {}", state.intervals.get(id));
+        trace!("main loop: allocating {}", state.intervals.get(id));
 
         #[cfg(debug_assertions)]
         {
@@ -237,15 +237,15 @@ pub(crate) fn run<F: Function>(
             reusable.computed_inactive = false;
         }
 
-        debug!("");
+        trace!("");
     }
 
-    if log_enabled!(Level::Debug) {
-        debug!("allocation results (in order):");
+    if log_enabled!(Level::Trace) {
+        trace!("allocation results (in order):");
         for int in state.intervals.virtuals.iter() {
-            debug!("{}", int);
+            trace!("{}", int);
         }
-        debug!("");
+        trace!("");
     }
 
     Ok((state.intervals, state.next_spill_slot.get()))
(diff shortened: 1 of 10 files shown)


Amanieu:
I would argue that this should be `trace!` since it is extremely verbose (lots of output for every function that is compiled).

bnjbvr (changes requested):
Thanks for doing this! Agreed with @Amanieu here, apart from the logs at the end of the `bt_main` top-level function, it'd be ok to have most, if not all, of these be `trace!` statements instead. This logging output really is intended for debugging purposes, and can be quite expensive to compute, e.g. when displaying a whole function (it's particularly slow to output a whole function in Cranelift).

pepyakin:
Ok, here is another take: I demoted everything, include those weren't touched by the original diff, to `trace!`. Exceptions are `bt_main.rs` and `lib.rs` where I put them into the `debug` bucket since they are once per regalloc run.

I am not sure if I can go further however: the code is pretty dense and I am not sure about frequencies. I am happy to address any direct feedback though.

pepyakin:
Kindly requesting another look )

cfallin (approved):
LGTM! Thanks, and sorry about the logging noise!

lib/src/linear_scan/assign_registers.rs
```diff
@@ -8,7 +8,7 @@ use crate::{
     VirtualReg, NUM_REG_CLASSES,
 };
 
-use log::{debug, info, log_enabled, trace, Level};
+use log::{trace, log_enabled, Level};
```
bnjbvr, line 11: If you don't mind, can you revert the `debug` -> `trace` changes in this file, and simply morph `info` to `debug` instead? I was using `debug` for low-level information that's useful to keep track of what's going on, vs `trace` vs very verbose mode, so the previous was rather more useful than it would be now. (also linear scan isn't used by default in wasmtime, so this shouldn't be a worry)

lib/src/linear_scan/resolve_moves.rs
```diff
@@ -7,7 +7,7 @@ use crate::{
     Function, RealReg, Reg, SpillSlot, TypedIxVec, VirtualReg, Writable,
 };
 
-use log::{debug, info, trace};
+use log::trace;
```
bnjbvr, line 10: and ditto in this file

bnjbvr (changes requested):
Sorry for getting back to this this late! I wasn't sure whether we wanted to keep on maintaining this version of regalloc, but seems like we do for a short while at least.

If you don't mind, can you please revert most of the changes in two linear scan files (see below) in a separate commit? And then I promise to merge this quickly :-)

pepyakin:
Hey @bnjbvr, I've reverted my changes for the two files you mentioned and changed two instances of `info!` into `debug!`. The change is in this commit https://github.com/bytecodealliance/regalloc.rs/pull/123/commits/90547f2a2ff0d8fd64c73c3d3320f3a91b2823f6. 

Besides that I've rebased the changes against main, hence the force push.

bnjbvr (approved):
LGTM, and sorry about the unexpected rustfmt changes / thanks for fixing it :)

pepyakin:
No worries, thanks!

PR #125, opened 2021-07-06
Bump dependencies + remove build warnings

bnjbvr:

PR #126, opened 2021-08-09
Fix release build

alexcrichton:
Currently this crate fails to build with `--release` due to a missing
method, and this commit fixes that! I believe this was an accidental
regression from #125.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -26,6 +26,8 @@ jobs:
       run: cargo build
     - name: Run tests
       run: cargo test --all --verbose
+    # make sure code builds in release mode as well
+    - run: cargo check --release
 
   # Lint dependency graph for security advisories, duplicate versions, and
   # incompatible licences.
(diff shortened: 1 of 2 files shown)


cfallin:
Eek, sorry for not catching that! Would you mind adding a release build to the CI config so we catch this in the future?

alexcrichton:
Sure!

bnjbvr (approved):

.github/workflows/rust.yml
```diff
@@ -26,6 +26,8 @@ jobs:
       run: cargo build
```
bnjbvr, line 26: Could we also just `cargo check` here?

alexcrichton: Technically this step isn't really that necessary since the `cargo test` below will do the same thing. Do you want me to remove this?

bnjbvr: Fair point, please do!

alexcrichton: Sure thing, should be updated now

cfallin (approved):
Thanks!

PR #130, opened 2021-11-19
Wrap one more block calling `show_with_rru` into `log_enabled!()`

bnjbvr:
Previous pull request did add log_enabled!() checks to make logging (via
show_with_rru) less costful, but one place was missing. This patches
fixes it.


cfallin (approved):
Good catch, thanks!

PR #131, opened 2021-11-22
Release 0.0.33

bnjbvr:
The log fix from #130 causes a large speedup when the logger implementation correctly implements `Log::enabled`, so I'm proposing to do a patch release here. Should be pretty straightforward to get in wasmtime as well, also it might not even be necessary, as this is a patch version and Cargo should pick it up automatically (? in theory at least).

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "regalloc"
-version = "0.0.32"
+version = "0.0.33"
 authors = ["The Regalloc.rs Developers"]
 edition = "2018"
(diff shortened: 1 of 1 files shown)


bjorn3:
All 0.0.x releases are considered incompatible by cargo AFAIK.

bnjbvr:
TIL! Well then, it's going to be a simple patch to write in wasmtime, I guess :-)

cfallin (approved):
LGTM! You're planning to do the `cargo publish` then? (I'm happy to if not for some reason)

bnjbvr:
> You're planning to do the cargo publish then?

Yep, done!

PR #139, opened 2022-01-19
Make `WritableBase` public and add a blanket implementation

fitzgen:
The blanket implementation is for any type that implements all of
`WritableBase`'s super traits.

This allows us to do things like create newtype wrappers for different register
classes (e.g. `Gpr` or `Xmm` on x86_64) and still use these newtype wrappers
with `Writable` (e.g. `Writable<Xmm>`).

r? @cfallin

### DIFF
--- a/lib/src/data_structures.rs
+++ b/lib/src/data_structures.rs
@@ -901,10 +901,6 @@ impl Reg {
 /// create a distinction, at the Rust type level, between a plain "register"
 /// and a "writable register".
 ///
-/// Only structs that implement the `WritableBase` trait can be wrapped with
-/// `Writable`. These are the Reg, RealReg and VirtualReg data structures only,
-/// since `WritableBase` is not exposed to end users.
-///
 /// Writable<..> can be used by the client to ensure that, internally, it only
 /// generates instructions that write to registers that should be written. The
 /// `InstRegUses` below, which must be implemented for every instruction,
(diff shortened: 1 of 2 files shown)


bnjbvr:
It's a bit unfortunate, because we really wanted to allow random types being used in `Writable`, at the point of writing. If you just made `WritableBase` public, would it suffice so that you can add the empty impl for your own defined data structures? (don't think that would conflict with rust's rules)

fitzgen:
I could do that, but why make people write an empty implementation?

(Why even have `WritableBase`, really? It is a bit of an anti-pattern IMO. It seems like its mostly a desire not to repeat `where T: Copy + Eq + etc...`. I didn't make this change though because it is more invasive.)

cfallin (approved):
Thanks! IMHO the blanket impl is fine since there isn't really anything custom that one would want to do (indeed, could do) if one were to implement it on the user side. I don't remember if there was any reason why we wanted to restrict what type could go into a `Writable`; the comment that this PR deletes makes it sound like there was a reason, but I'm more inclined to agree with @bnjbvr here that there shouldn't have been.

Anyway, r+ from me...

bnjbvr:
Could it be that one can add the wrapper to a type that's already wrapped, and end up with `Writable<Writable<T>>`? In any case, change looks fine to me, and this code is doomed anyways :)

fitzgen:
Yeah there is nothing stopping someone from doing `Writable<Writable<T>>` except perhaps code review :)

fitzgen:
Anyways, I don't have merge permissions for this repo. Could someone else merge this and cut a point release, please? Thanks!

PR #144, opened 2022-10-03
Deprecate this crate

bnjbvr:
We are deprecating this crate, since regalloc2 is more modern, maintained and expected to be faster both in compile times and have better generated code quality.

@cfallin If you agree with this, once we've merged this, I'll archive this repository and disable pull requests/issues. If there's any interesting issues that also makes sense for regalloc2, we can re-open them in regalloc2's issue tracker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 regalloc.rs
 ===
 
-⚠ This project has been deprecated in favor of the [regalloc2](https://github.com/cfallin/regalloc2), which is maintained and expected to be faster. No security updates will be made, and contributions will be disabled.
+⚠ This project has been deprecated in favor of the [regalloc2](https://github.com/bytecodealliance/regalloc2), which is maintained and expected to be faster. No security updates will be made, and contributions will be disabled.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 regalloc.rs
 ===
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,10 @@
 regalloc.rs
 ===
 
+⚠ This project has been deprecated in favor of the [regalloc2](https://github.com/cfallin/regalloc2), which is maintained and expected to be faster. No security updates will be made, and contributions will be disabled.
```
bjorn3, line 4: Is regalloc2 not under the bytecodealliance organization?

bnjbvr: Aouch, good catch, thanks!

bnjbvr, line 4:
```suggestion
⚠ This project has been deprecated in favor of the [regalloc2](https://github.com/bytecodealliance/regalloc2), which is maintained and expected to be faster. No security updates will be made, and contributions will be disabled.
```

cfallin (approved):
Thanks @bnjbvr -- and thanks to regalloc.rs for carrying Cranelift well for several years!

cfallin:
cargo-deny is failing because the crate `unicode-ident` (pulled in via some proc-macro helpers, then serde) added an "and Unicode license" qualifier to its license field. Since this crate is no longer used in downstream contexts that have strict license requirements, and because such a license would likely not affect the actual compiled code, and because the tree today is the same as it was yesterday (this PR doesn't introduce any changes to dependencies), I'm going to go ahead and merge this against the CI failure.

